PR #35, opened 2024-03-05
Refactoring `Internal` module for arithmetic circuits

iokasimov:
This is initial PR in order to get to know compiler internals.

It introduces some basic optics for convenient work with stateful records.
Also it removes unnecessary `take` in `apply` expression.

It's up to you if you don't want to accept it since it's pretty subjective.


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -117,51 +114,40 @@ type Constraint a = SomePolynomial a
...
-        n   = length xs
-    in r
-    {
-        acInput = drop n ins,
-        acWitness = acWitness r . (fromList (zip (take n ins) xs) `union`)
-    })
+apply xs = do
+    inputs <- zoom #acInput get
+    zoom #acInput . put $ drop (length xs) inputs
+    zoom #acWitness . modify $ (. (union . fromList $ zip inputs xs))
```
TurtlePU, line 150: Flipping `union` arguments would make it a little more readable: `(. union (fromList $ zip inputs xs))`

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -92,21 +94,16 @@ toVar srcs c = fromBinary $ castBits $ binaryExpansion ex
...
+newVariableWithSource srcs con = toVar srcs . con . fst <$> do
+    zoom #acRNG $ get >>= traverse put . randomR (0, order @VarField - 1)
 
-addVariable :: Integer -> State (ArithmeticCircuit a) ()
-addVariable x = modify (\r -> r { acOutput = x, acVarOrder = insert (length (acVarOrder r), x) x (acVarOrder r)})
+addVariable :: Integer -> State (ArithmeticCircuit a) Integer
+addVariable x = do
+    zoom #acOutput $ put x
+    zoom #acVarOrder . modify
+        $ insert <$> (,x) . length <*> const x <*> id
```
TurtlePU, line 105: I would rather introduce a new variable for `acVarOrder` argument, it would be a little easier to read

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -117,51 +114,40 @@ type Constraint a = SomePolynomial a
...
+forceZero = zoom #acOutput get >>= constraint . var
 
 -- | Adds a new variable assignment to the arithmetic circuit.
 -- TODO: forbid reassignment of variables
-assignment :: forall a . (Map Integer a -> a) -> State (ArithmeticCircuit a) ()
-assignment f = modify $ \r -> r { acWitness = (insert (acOutput r) =<< f) . acWitness r }
+assignment :: (Map Integer a -> a) -> State (ArithmeticCircuit a) ()
+assignment f = do
+    i <- insert <$> zoom #acOutput get
+    zoom #acWitness . modify $ (.) (f >>= i)
```
TurtlePU, line 128: As `>>=` here is from `(e ->)` monad, but it is used inside do-notation for `State`, I would once again unfold the usage of `(e ->)` monad in order to not confuse the reader.

TurtlePU (approved):

zkFold (approved):

PR #55, opened 2024-03-19
Add naturals

TurtlePU:
This is a big one.
* Made `Semiring` into a class with a `FromConstant Natural` constraint;
* Added an initial semiring instance -- `instance Semiring Natural`;
* Migrated most functions and types from `Integer` to `Natural` as this makes much more sense.

1. Some migrations might be unnecessary, in this case I could change it back.
2. Maybe, we could avoid specifying the concrete type of variables in polynomials until where this is actually needed. Most of the time, there is no substantial difference between Naturals and Integers.

### DIFF
--- a/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
+++ b/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
@@ -44,6 +44,7 @@ lt :: Poly c -> c
 lt (P cs) = last cs
 
 deg :: Poly c -> Integer
+-- | Degree of zero polynomial is `-1`
 deg (P cs) = fromIntegral (length cs) - 1
 
 scaleP :: Ring c => c -> Natural -> Poly c -> Poly c
@@ -56,6 +57,7 @@ qr a b = go a b zero
             where
                 c = lt x / lt y
                 n = fromIntegral (deg x - deg y)
+                -- ^ if `deg x < deg y`, `n` is not evaluated, so this would not error out
                 x' = x - scaleP c n y
                 q' = q + scaleP c n one
 
--- a/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
+++ b/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
@@ -3,11 +3,12 @@
 
 module ZkFold.Base.Algebra.Polynomials.Univariate where
 
-import           Prelude                           hiding (Num(..), (/), (^), sum, product, length, replicate, take, drop)
-import           Test.QuickCheck                   (Arbitrary(..))
+import           Numeric.Natural                 (Natural)
+import           Prelude                         hiding (Num (..), drop, length, product, replicate, sum, take, (/), (^))
+import           Test.QuickCheck                 (Arbitrary (..))
 
 import           ZkFold.Base.Algebra.Basic.Class
-import           ZkFold.Prelude                    (replicate, length, take, drop, zipWithDefault)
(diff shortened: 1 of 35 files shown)


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
```diff
@@ -104,19 +105,22 @@ class Monad m => MonadBlueprint i a m | m -> i, m -> a where
...
     newConstrained
-        :: NewConstraint Integer a
-        -> Witness Integer a
-        -> State (ArithmeticCircuit a) Integer
+        :: NewConstraint Natural a
+        -> Witness Natural a
+        -> State (ArithmeticCircuit a) Natural
     newConstrained new witness = do
-        let s = sources witness `Set.difference` sources (`new` (-1))
+        let ws = sources witness
```
iokasimov, line 120: @TurtlePU Can you elaborate here a little bit please?

TurtlePU: `s` is meant to be a set of variables used in a witness not present in a constraint. To compute the set of variables not present in a constraint, we first have to pass a "new variable" which we are generating, but here it is enough to pass a variable which definitely would not be present in a witness

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -43,9 +44,9 @@ lt :: Poly c -> c
 lt (P cs) = last cs
 
 deg :: Poly c -> Integer
```
vks4git, line 46: Was Integer left here on purpose?

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -54,7 +55,7 @@ qr a b = go a b zero
         go x y q = if deg x < deg y then (q, x) else go x' y q'
             where
                 c = lt x / lt y
-                n = deg x - deg y
+                n = fromIntegral (deg x - deg y)
```
vks4git, line 58: What happens if the difference is negative and hence is not a Natural?

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -55,25 +53,27 @@ greedySplitAt n (x : xs) =
...
+    x + y = fromConstant $ toNatural x + toNatural y
 
 instance (Finite p, KnownNat n) => AdditiveMonoid (UInt n (Zp p)) where
-    zero = fromConstant (0 :: Integer)
+    zero = fromConstant (0 :: Natural)
 
 instance (Finite p, KnownNat n) => AdditiveGroup (UInt n (Zp p)) where
-    x - y = fromConstant $ toInteger x - toInteger y
-    negate = fromConstant . negate . toInteger
+    x - y = fromConstant $ toNatural x - toNatural y
```
vks4git, line 67: Again, what happens if x < y?

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
TurtlePU: If the difference is negative (`deg x < deg y`), `n` is not evaluated at all, see above

vks4git (approved):

iokasimov (approved):

vlasin (approved):

PR #61, opened 2024-04-01
Rewrite Cardano's `Transaction` and `Address` implementations

iokasimov:
Changes needed in [Payment module in P2P](https://github.com/zkFold/p2p-onramp/blob/main/src/ZkFold/P2P/Payment.hs)
Related task: https://github.com/zkFold/zkfold-base/issues/10


src/ZkFold/Symbolic/Cardano/Types.hs
```diff
@@ -1,13 +1,101 @@
...
+
+deriving instance
+    ( KnownNat n
+    , Arithmetizable i a
+    , Arithmetizable i (UInt 64 a)
+    , Arithmetizable i (ByteString 8 a)
+    , Arithmetizable i (ByteString 15 a)
+    ) => Arithmetizable i (Value n a)
+
+newtype Input datum a = Input (OutputRef a, (a, datum a))
```
vlasin, line 56: Inputs should be pairs of (OutputRef, Output).

src/ZkFold/Symbolic/Cardano/Types.hs
```diff
@@ -1,13 +1,101 @@
...
+txInputs :: Transaction inputs rinputs outputs tokens datum a -> Vector inputs (Input datum a)
+txInputs (Transaction (_, (is, _))) = is
+
+txOutputs :: Transaction inputs rinputs outputs tokens datum a -> Vector outputs (Output tokens a)
+txOutputs (Transaction (_, (_, (os, _)))) = os
+
+newtype TxId a = TxId a
+    deriving (Eq, Arithmetizable i)
+
+newtype Value n a = Value (Vector n (ByteString 8 a, (ByteString 15 a, UInt 64 a)))
```
vlasin, line 45: 8 and 15 are the number of bits in `CurrencySymbol` and `TokenName`. respectively. It should be 224 bit for the `CurrencySymbol`, and either 256 or 512 bit for the `TokenName` (please check).

src/ZkFold/Symbolic/Cardano/Types.hs
```diff
@@ -1,13 +1,101 @@
...
+    ) => Arithmetizable i (Output tokens a)
+
+newtype OutputRef a = OutputRef (TxId a, UInt 32 a)
+    deriving Eq
+
+deriving instance
+    (Arithmetizable i a, Arithmetizable i (UInt 32 a))
+    => Arithmetizable i (OutputRef a)
+
+newtype Address a = Address (ByteString 3 a, (ByteString 28 a, ByteString 28 a))
```
vlasin, line 89: In `ByteString n a`, `n` is the number of bits! Both credentials are 224-bit.

iokasimov: @vlasin Sorry, I mistakenly assumed without asking that it's a byte string length.

src/ZkFold/Symbolic/Cardano/Types.hs
iokasimov: @vlasin `TokenName` should be no longer than 32 bytes so it's `256`.

vlasin (approved):

PR #62, opened 2024-04-01
Bitwise operations on bytestrings

vks4git:


src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -0,0 +1,130 @@
...
+import           ZkFold.Base.Algebra.Basic.Class
+import           ZkFold.Base.Algebra.Basic.Field                           (Zp)
+import           ZkFold.Prelude                                            (replicate)
+import           ZkFold.Symbolic.Compiler
+import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.Combinators    (expansion, horner)
+import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.MonadBlueprint (ClosedPoly)
+import           ZkFold.Symbolic.Data.Combinators
+import           ZkFold.Symbolic.Data.UInt
+
+newtype ByteString (n :: Natural) a = ByteString (UInt n a)
```
TurtlePU, line 27: `UInt n a` needs some additional space to perform integer arithmetic, `ByteString`s could be stored more efficiently

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -0,0 +1,130 @@
...
+instance (Finite p, KnownNat n) => ToConstant (ByteString n (Zp p)) Natural where
+    toConstant (ByteString uint) = toConstant uint
+
+deriving newtype instance (Finite p, KnownNat n) => Arbitrary (ByteString n (Zp p))
+deriving newtype instance (FromConstant Natural a, Finite a, AdditiveMonoid a, KnownNat n) => FromConstant Natural (ByteString n a)
+deriving newtype instance (FromConstant Natural a, Finite a, AdditiveMonoid a, KnownNat n) => FromConstant Integer (ByteString n a)
+
+--------------------------------------------------------------------------------
+
+instance (Finite p, KnownNat n) => AdditiveSemigroup (ByteString n (Zp p)) where
```
TurtlePU, line 41: I would implement `BoolType` for `ByteString` instead of algebraic classes

tests/Main.hs
```diff
@@ -14,6 +14,7 @@ import           Tests.Permutations                          (specPermutations)
 import           Tests.Plonk                                 (specPlonk)
 import           Tests.Scripts.LockedByTxId                  (specLockedByTxId)
 import           Tests.UInt                                  (specUInt)
+import           Tests.ByteString (specByteString)
```
TurtlePU, line 17: Could you use a stylish-haskell formatter to format this line? We have configs for VS Code and NeoVim in this repository, but could support more editors if needed

examples/Examples/ByteString.hs
```diff
@@ -0,0 +1,38 @@
...
+import           ZkFold.Symbolic.Compiler                    (ArithmeticCircuit, compileIO)
+import           ZkFold.Symbolic.Data.ByteString
+
+exampleByteStringAnd :: forall n . KnownNat n => IO ()
+exampleByteStringAnd = makeExample @n "*" "and" (*)
+
+exampleByteStringOr :: forall n . KnownNat n => IO ()
+exampleByteStringOr = makeExample @n "+" "or" (+)
+
+type Binary a = a -> a -> a
```
vlasin, line 29: Should we move this definition to the lib?

src/ZkFold/Symbolic/Data/ByteString.hs
vlasin: I guess it is in the task description. But I agree that if we can pack more tightly, then we should.

examples/Examples/ByteString.hs
TurtlePU: I guess not as it is only used for brevity and only in examples
Maybe later, when we see the need for a general approach to preparing examples

tests/Main.hs
vks4git: Just missed this file. Of course, I always run stylish-haskell :)

vlasin (approved):

TurtlePU (commented):
ok

PR #70, opened 2024-04-05
Redesign Scale

TurtlePU:
* Added documentation with laws to all of the algebraic classes
* Removed non-lawful instances (`AdditiveGroup Natural`, `BinaryExpansion Integer`, `MultiplicativeGroup (Poly ...)`), applied fixes accordingly
* Moved binary-related scaling and exponentiation to functions from instances
* Removed all-too-broad instances of `Scale` and `Exponent` to allow for better specialized instances
* Made `Scale Natural`, `Scale Integer`, `Exponent Natural` and `Exponent Integer` superclass constraints where applicable
* Removed newtypes in `Algebra/Basic/Scale` as they are not needed anymore


src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -1,81 +1,321 @@
...
+    --
+    -- [Subtraction] @x - x == zero@
+    -- [Agreement] @x - y == x + negate y@
+    (-) :: a -> a -> a
+    x - y = x + negate y
+
+    -- | Inverse in an abelian group. The following should hold:
+    --
+    -- [Negative] @x + negate x == zero@
+    -- [Agreement] @invert x == zero - x@
```
vlasin, line 232: typo here

src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -220,26 +475,36 @@ instance BinaryExpansion Bool where
...
     one = repeat one
 
 instance MultiplicativeGroup a => MultiplicativeGroup [a] where
     invert = map invert
 
+instance AdditiveSemigroup a => AdditiveSemigroup [a] where
+    (+) = zipWith (+)
+
+instance Scale b a => Scale b [a] where
+    scale = map . scale
```
vks4git, line 500: How about generalising it to Functor? This way polynomials which have Vectors under the hood will get Scale almost for free.
```suggestion
instance Functor f => Scale b a => Scale b (f a) where
    scale = fmap . scale
```

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -33,6 +33,9 @@ instance (Ring c, Eq c) => AdditiveSemigroup (Poly c) where
         lPadded = l V.++ V.replicate (len P.- V.length l) zero
         rPadded = r V.++ V.replicate (len P.- V.length r) zero
 
+instance Scale c' c => Scale c' (Poly c) where
+    scale c' (P v) = P $ V.map (scale c') v
```
vks4git, line 37: This could be made a bit prettier with Scale defined on Functors
```suggestion
    scale c' (P v) = P $ scale c' v
```

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
```diff
@@ -21,25 +21,23 @@ import           Data.Set                                            (Set)
...
 import           ZkFold.Prelude                                      (replicate)
 import qualified ZkFold.Symbolic.Compiler.ArithmeticCircuit.Internal as I
 import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.Internal hiding (constraint)
-import           ZkFold.Symbolic.Data.Bool                           (Bool (..), BoolType (..))
+import           ZkFold.Symbolic.Data.Bool                           (Bool (..))
 import           ZkFold.Symbolic.Data.Conditional                    (Conditional (..))
 import           ZkFold.Symbolic.Data.Eq                             (Eq (..))
 
-type WitnessField x a = (Algebra x a, FiniteField x, BinaryExpansion x,
+type WitnessField a x = (Algebra a x, FiniteField x, BinaryExpansion x,
```
vks4git, line 34: Why was this change made?

src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -249,24 +514,30 @@ instance Ring a => Ring [a]
...
 instance MultiplicativeMonoid a => MultiplicativeMonoid (p -> a) where
     one = const one
 
 instance MultiplicativeGroup a => MultiplicativeGroup (p -> a) where
     invert = fmap invert
 
+instance AdditiveSemigroup a => AdditiveSemigroup (p -> a) where
+    p1 + p2 = \x -> p1 x + p2 x
+
+instance Scale b a => Scale b (p -> a) where
```
vks4git, line 532: I suppose we would get this for free if Scale is defined on Functors

src/ZkFold/Base/Algebra/Basic/Class.hs
TurtlePU: This is a great idea, however, once we make an `ArithmeticCircuit` into a functor, this instance would do the wrong thing for it.
(But maybe this is an argument against making `ArithmeticCircuit` a functor, @vlasin, @iokasimov, what do you think?)

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
TurtlePU: The idea is that the last type parameter should be the "main" one (in this case -- the carrier of an algebraic structure)
But I guess there is no problem in switching places

src/ZkFold/Base/Algebra/Basic/Class.hs
TurtlePU: Yeah, definitely

src/ZkFold/Base/Algebra/Basic/Class.hs
TurtlePU: fixed, thank you

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
vlasin: I think this one is fine. I agree with @TurtlePU's logic.

I was thinking about `Exponent b a`, though. The logic was that the order of type arguments is the same as in the mathematical notation of the related expressions (`Scale b a`, but `Exponent a b`). What do you think? Is it better after the change of before?

src/ZkFold/Base/Algebra/Basic/Class.hs
vlasin: Yes, for now, it is not clear if `ArithmeticCicruit` should be a functor. I like the proposed generalization.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
TurtlePU: Oh, actually `Exponent a b` makes more sense
When I myself was preparing this PR, I would mess the order up sometimes, so I guess `Exponent a b` is more intuitive, will change to that

TurtlePU: @vlasin, changed the order to `Exponent a b`

src/ZkFold/Base/Algebra/Basic/Class.hs
iokasimov: @vlasin @TurtlePU Probably it's possible to make it a `Functor` instance if we make `acWitness` a continuation.

TurtlePU: @iokasimov, it is, but then a `Scale` instance for functors would silently do the wrong thing for it if one doesn't import the correct import from Instances module

Issue #65, opened 2024-04-01
Optimize `bool` and constant embedding


vlasin:
- [x] Optimize `bool` arithmetization to three constraints
- [ ] Embed constants without creating a constraint

echatav:
With regard to constant embedding, I verified that regardless of the integer chosen at random, embedding creates 1 variable and 1 constraint:
```
Example: constant 6


Compiling the script...

Number of constraints: 1
Number of variables: 1
Script saved: compiled_scripts/constant.json



zkfold-base> Test suite zkfold-base-examples passed
Completed 2 action(s).
hephaestus@Eitans-MacBook-Pro zkfold-base % stack build zkfold-base:zkfold-base-examples
zkfold-base> test (suite: zkfold-base-examples)


Example: constant 93


Compiling the script...

Number of constraints: 1
Number of variables: 1
Script saved: compiled_scripts/constant.json
```
The code for `fromConstant` calls `embed` which is defined as:

```Haskell
embed :: Arithmetic a => a -> ArithmeticCircuit a
embed x = circuit $ newAssigned $ const (fromConstant x)
```

Finally, `newAssigned` is defined as:
```Haskell
    newAssigned :: MonadBluePrint i a m => ClosedPoly i a -> m i
    newAssigned p = newConstrained (\x i -> p x - x i) p
```

From here, it's evident where the 1 constraint comes from. And it's not clear to me how to correctly excise this constraint, assuming that's possible.

echatav:
On the other hand, it's not clear that the definition of `bool` is even calling `fromConstant`. It doesn't appear to be:

```Haskell
instance SymbolicData a x => Conditional (Bool (ArithmeticCircuit a)) x where
    bool brFalse brTrue (Bool b) =
        let f' = pieces brFalse
            t' = pieces brTrue
        in restore $ zipWith (\f t -> b * (t - f) + f) f' t'
```

I must ask, on what basis are we convinced that `bool` can be arithmetized with 2 constaints, not 3?

echatav:
We should also note that the number of variables and constraints for `bool` is highly dependent on the symbolic datatype it operates on:
```
Number of constraints: 0
Number of variables: 0
Script saved: compiled_scripts/conditional0.json

Example: conditional


Compiling the script...

Number of constraints: 3
Number of variables: 4
Script saved: compiled_scripts/conditional.json

Example: conditional on pair


Compiling the script...

Number of constraints: 4
Number of variables: 5
[...]

PR #80, opened 2024-04-11
Vks4git hash

vks4git:


src/ZkFold/Symbolic/Data/Bool.hs
```diff
@@ -50,6 +54,8 @@ instance Field x => BoolType (Bool x) where
 
     (||) (Bool b1) (Bool b2) = Bool $ b1 + b2 - b1 * b2
 
+    xor (Bool b1) (Bool b2) = Bool $ b1 + b2 - (b1 * b2 + b1 * b2)
```
vlasin, line 57: In #65, we aim to make constants free. In particular, that would mean that `(b1 * b2 + b1 * b2) == 2 * b1 * b2`. So, the whole `xor` function can be implemented with one Plonk constraint.

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -1,60 +1,191 @@
...
+
+    rotateBitsL :: a -> Natural -> a
+    rotateBitsL a s = rotateBits a (Haskell.fromIntegral s)
+
+    rotateBitsR :: a -> Natural -> a
+    rotateBitsR a s = rotateBits a (negate . Haskell.fromIntegral $ s)
+
+
+-- | Describes types which can be split into words of equal size.
+-- Parameters have to be of different types as ByteString store their lengths on type level and hence after splitting they chage types.
```
vlasin, line 84: typo "change"

src/ZkFold/Symbolic/Data/Bool.hs
vlasin: But since it is not implemented yet, maybe we could leave it like that.

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -133,7 +468,88 @@ instance (Arithmetic a, KnownNat n) => BoolType (ByteString n (ArithmeticCircuit
...
+        solve = do
+            bits <- toBits @n @a x xs
+            let words = chunksOf (Haskell.fromIntegral $ getNatural @wordSize) bits
+            wordsBits <- mapM (fromBits @wordSize @a) words
+            pure $ Haskell.concat wordsBits
+
+instance
+  ( KnownNat m
+  , KnownNat n
+  , m <= n
```
TurtlePU, line 504: Do you need this constraint? Logically it is implied from `Mod n m ~ 0` (except if `n == 0`)

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -86,15 +331,104 @@ instance (Arithmetic a, KnownNat n) => Arithmetizable a (ByteString n (Arithmeti
 
     typeSize = minNumberOfRegisters @a @n
 
-binOp
-    :: forall a n .
-       Arithmetic a
+
+-- TODO: I really don't like that summation is implemented here and in UInt. Can we do something about it?
```
TurtlePU, line 335: And I don't like that `ByteString` is used for addition as
1. An efficient implementation would require `ByteString` to reserve space just as in `UInt` (and why not use `UInt` then)
2. The current implementation which uses a bitwise summation is infeasible

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -86,15 +331,104 @@ instance (Arithmetic a, KnownNat n) => Arithmetizable a (ByteString n (Arithmeti
...
+-- Converting ByteStrings to UInt and back will flood the circuit with new constraints because of different bit layouts in these types.
```
TurtlePU, line 336: Well, this is actually not a problem because
1. The cast needed in SHA-2 algorithms is between `ByteString 32` and `UInt 32` (or between `ByteString 64` and `UInt 64`), which is a no-op in the fields we use because they both need a single register
2. If someone needs a cast for big `n`, this would not be a problem anyway once we add range checks

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -133,7 +468,88 @@ instance (Arithmetic a, KnownNat n) => BoolType (ByteString n (ArithmeticCircuit
...
+
 
-    l && r = binOp l r (\i j x -> x i * x j)
+instance
+  ( KnownNat wordSize
+  , KnownNat n
+  , wordSize <= n
+  , 1 <= wordSize
+  , 1 <= n
+  , Mod n wordSize ~ 0
```
TurtlePU, line 484: Same here: `Mod n wordSize ~ 0` logically implies `1 <= wordSize`. And, together with `1 <= n` -- that `wordSize <= n`

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -75,8 +206,122 @@ instance (KnownNat p, KnownNat n) => BoolType (ByteString n (Zp p)) where
...
+
+        wordSize :: Natural
+        wordSize = getNatural @wordSize
+
+        natWords :: [Natural]
+        natWords = unfoldr (toBase (2 Haskell.^ wordSize)) asNat <> Haskell.repeat (fromConstant @Natural 0)
+
+-- | Unfortunately, Haskell does not support dependent types yet,
+-- so we have no possibility to infer the exact type of the result
+-- (the list can contain an arbitrary number of words).
```
TurtlePU, line 245: Maybe, a `Vector k a` can be used instead of `[a]`, then?

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -1,60 +1,191 @@
...
+-- | Convert a number into @base@-ary system.
+--
+toBase :: FromConstant Natural a => Natural -> Natural -> Maybe (a, Natural)
+toBase _ 0    = Nothing
+toBase base b = let (d, m) = b `divMod` base in Just (fromConstant m, d)
+
 
 instance (FromConstant Natural a, Finite a, KnownNat n) => FromConstant Integer (ByteString n a) where
-    fromConstant = fromConstant . naturalFromInteger . (`Haskell.mod` (2 ^ getNatural @n))
+    fromConstant = fromConstant . naturalFromInteger . (`Haskell.mod` (2 Haskell.^ getNatural @n))
```
TurtlePU, line 137: `^` for Integers and Naturals is a standard Haskell exponentiation already

src/ZkFold/Symbolic/Algorithms/Hash/SHA2.hs
```diff
@@ -0,0 +1,368 @@
...
+                                                                 sha512_224InitialHashes, sha512_256InitialHashes,
+                                                                 word32RoundConstants, word64RoundConstants)
+import           ZkFold.Symbolic.Data.Bool                      (BoolType (..))
+import           ZkFold.Symbolic.Data.ByteString                (ByteString (..), Concat (..), Extend (..),
+                                                                 ShiftBits (..), ToWords (..), Truncate (..))
+
+-- | SHA2 is a family of hashing functions with almost identical implementations but different constants and parameters.
+-- This class links these varying parts with the appropriate algorithm.
+--
+class AlgorithmSetup (algorithm :: Symbol) a where
```
TurtlePU, line 35: This is beautiful!

src/ZkFold/Symbolic/Algorithms/Hash/SHA2.hs
```diff
@@ -0,0 +1,368 @@
...
+            toWords' 0 = []
+            toWords' n = fromConstant (n `mod` base) : toWords' (n `div` base)
+
+            base :: Natural
+            base = 2 P.^ (value @n)
+
+
+-- | Constraints required for a SHA2 of a Natural number.
+--
+type SHA2N algorithm element =
```
TurtlePU, line 225: @vlasin, and this is the reason why I once suggested to introduce a `class Symbolic a` with instances in `Zp` and `ArithmeticCircuit`

src/ZkFold/Symbolic/Data/ByteString.hs
vks4git: Logically, yes. But Haskell is unable to infer `n <= n + a `or `Mod (a * n) n ~ 0`.

src/ZkFold/Symbolic/Data/ByteString.hs
vks4git: Are these vectors already implemented somewhere?

src/ZkFold/Symbolic/Data/ByteString.hs
vks4git: Oh wait, your question is about a different thing. Yeah, probably it is safe to remove this

src/ZkFold/Symbolic/Data/ByteString.hs
vks4git: Oh, but then it would make testing impossible. There should be some loophole to calculate hashes of byte strings with lengths unknown at compile time

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -1,69 +1,206 @@
...
+
+            d :: Natural
+            d = nat `shiftR` intS
+
 
 instance (KnownNat p, KnownNat n) => BoolType (ByteString n (Zp p)) where
     false = fromConstant (0 :: Natural)
 
-    -- | A ByteString with all bits set to 1 is the unity for bitwise and.
+    -- | A ByteString with all bits set from 1 is the unity for bitwise and.
```
TurtlePU, line 199: Typo (after renaming `to -> from`, I guess)

vks4git: oops

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -63,7 +200,7 @@ instance (KnownNat p, KnownNat n) => BoolType (ByteString n (Zp p)) where
     true = not false
 
     -- | bitwise not.
-    -- @Data.Bits.complement@ is undefined for @Natural@, we have to flip bits manually.
+    -- @Data.Bits.complement@ is undefined for @Natural@, we have from flip bits manually.
```
TurtlePU, line 203: Another one here

TurtlePU (dismissed):

src/ZkFold/Symbolic/Data/Combinators.hs
```diff
@@ -13,6 +15,12 @@ import qualified Prelude                         as Haskell
 
 import           ZkFold.Base.Algebra.Basic.Class
 
+-- | A class for isomorphic types.
+-- The @Iso b a@ context ensures that transformations in both directions are defined
+--
+class Iso b a => Iso a b where
```
iokasimov, line 21: @vks4git Shouldn't there also be `into` method since you need a reverse morphism?

TurtlePU: It is implied by the superclass constraint

vks4git: Nope, `Iso b a` is listed as context, it will have the inverse mapping. You won't be able to define `Iso a b` without defining `Iso b a`, GHC won't compile such code

ghci> import ZkFold.Symbolic.Data.Combinators 
ghci> {instance Iso Bool (Maybe ()) where; from True = Just (); from False = Nothing}

<interactive>:2:11: error: [GHC-39999]
    • No instance for ‘Iso (Maybe ()) Bool’
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Iso Bool (Maybe ())’

TurtlePU (dismissed):

PR #83, opened 2024-04-12
Refactor polynomial modules

iokasimov:
Related issue: https://github.com/zkFold/zkfold-base/issues/17
I split this task into several PR's so it should be easier to review them.


zkfold-base.cabal
```diff
@@ -87,9 +87,7 @@ library
       ZkFold.Base.Algebra.EllipticCurve.Class
       ZkFold.Base.Algebra.Polynomials.Multivariate
       ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
-      ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial.Class
       ZkFold.Base.Algebra.Polynomials.Multivariate.Polynomial
-      ZkFold.Base.Algebra.Polynomials.Multivariate.Polynomial.Class
```
TurtlePU, line 92: But the file is not deleted, we need to delete it, maybe?

TurtlePU (approved):

PR #92, opened 2024-04-18
vector spaces

echatav:
In order to provide a theoretically sound basis for "symbolic data" let's start with vector spaces.


echatav:
Closing for now in favor of #107

PR #93, opened 2024-04-18
Refactor polynomial modules more

iokasimov:
Another round of polynomials refactoring


src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -38,15 +39,20 @@ instance Polynomial c i j => FromPolynomial c i j m [(c, M i j m)] where
     fromPolynomial = id
 
 class Polynomial c i j => ToPolynomial c i j m p where
-    toPolynomial   :: [(c, M i j m)] -> Maybe p
+    toPolynomial   :: [(c, M i j m)] -> p
```
TurtlePU, line 42: The motivation behind `Maybe` was that this operation is fallible e.g. because accumulated powers become too large. Do we throw an exception here now instead?

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -38,15 +39,20 @@ instance Polynomial c i j => FromPolynomial c i j m [(c, M i j m)] where
...
 
 instance (Polynomial c i j) => ToPolynomial c i j m [(c, M i j m)] where
-    toPolynomial   = Just . filter (\(c, _) -> c /= zero)
+    toPolynomial   = filter (\(c, _) -> c /= zero)
 
 -- | Polynomial type
 newtype P c i j m p = P p
     deriving (Generic, NFData, FromJSON, ToJSON)
 
+instance IsList (P c i j (Map i j) [(c, M i j (Map i j))]) where
```
TurtlePU, line 51: I think we can make it even more convenient by generalizing this to arbitrary `FromPolynomial c i j m p, ToPolynomial c i j m p`

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -57,7 +63,9 @@ instance (FromPolynomial c i j m p, FromMonomial i j m) => Eq (P c i j m p) wher
 
 -- TODO: this assumes sorted monomials! Needs fixing.
 instance (FromPolynomial c i j m p, FromMonomial i j m) => Ord (P c i j m p) where
-    compare (P l) (P r) = compare (map snd $ fromPolynomial @c @i @j @m l) (map snd $ fromPolynomial @c @i @j @m r)
+    compare (P l) (P r) = compare
```
TurtlePU, line 66: We can make this even more readable by using `Data.Function.on`

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
```diff
@@ -28,35 +40,78 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
...
+    Just i  -> i
+    Nothing -> error "mapVar: something went wrong"
+
+mapVarPolynomial :: [Natural] -> Polynomial' c -> Polynomial' c
+mapVarPolynomial vars (P ms) = P $ second (mapVarMonomial vars) <$> ms
+
+mapVarPolynomials :: [Natural] -> [Polynomial' c] -> [Polynomial' c]
+mapVarPolynomials vars = map (mapVarPolynomial vars)
+
+removeConstantVariable :: (Eq c, Field c) => Polynomial' c -> Polynomial' c
```
TurtlePU, line 115: This looks like a really special case, maybe we can move it closer to the place of usage?
(Also this can be replaced altogether by using `evalPolynomial` with a suitable substitution function)

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
```diff
@@ -28,35 +40,78 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
...
+
+mapVar :: [Natural] -> Natural -> Natural
+mapVar vars x = case x `elemIndex` vars of
+    Just i  -> i
+    Nothing -> error "mapVar: something went wrong"
+
+mapVarPolynomial :: [Natural] -> Polynomial' c -> Polynomial' c
+mapVarPolynomial vars (P ms) = P $ second (mapVarMonomial vars) <$> ms
+
+mapVarPolynomials :: [Natural] -> [Polynomial' c] -> [Polynomial' c]
```
TurtlePU, line 112: I think we can safely remove this function

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
```diff
@@ -28,35 +40,78 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
...
+
+mapVarMonomial :: [Natural] -> Monomial' -> Monomial'
+mapVarMonomial vars (M as) = M $ mapKeys (mapVar vars) as
+
+mapVar :: [Natural] -> Natural -> Natural
+mapVar vars x = case x `elemIndex` vars of
+    Just i  -> i
+    Nothing -> error "mapVar: something went wrong"
+
+mapVarPolynomial :: [Natural] -> Polynomial' c -> Polynomial' c
```
TurtlePU, line 109: Maybe we can generalize this to `(i -> i') -> P c i j m p -> P c i' j m p'`?
(And once again, this can be replaced by a call to `evalPolynomial`, because this is like an "fmap for Polynomial monad")

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
```diff
@@ -28,35 +40,78 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
...
+    $ fromPolynomial @c @i @j @m @p p
+        <&> (\(c, m) -> scale c $ evalMonomial f m)
 
-variables :: forall c i j m p . (FromMonomial i j m, FromPolynomial c i j m p) => P c i j m p -> [i]
-variables (P p) = nubOrd $ concatMap (\(_, M m) -> keys (fromMonomial @i @j @m m)) $ fromPolynomial @c @i @j @m @p p
+variables :: forall c i j m p .
+    FromMonomial i j m =>
+    FromPolynomial c i j m p =>
+    P c i j m p -> [i]
+variables (P p) = nubOrd
```
TurtlePU, line 89: I do similar thing using `Sources` newtype from `MonadBlueprint` module, maybe we can improve code reusage by using it (and moving it from `MonadBlueprint` somewhere else)?

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
iokasimov: @TurtlePU There is only one instance of `ToPolynomial`, we just add `Just` there. No I think that it's better to remove them, they add an unnecessary level of complexity.

echatav: If there's only 1 instance, why bother with a typeclass?

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
iokasimov: @TurtlePU I moved it here since we have to unpack `P` and `M`. So it's kinda internal stuff.

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
iokasimov: @TurtlePU But we use it in `Plonk` modules.

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
iokasimov: @TurtlePU Probably, but they have different types. in case of `sources` it's CPS encoding for witnesses.

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
iokasimov: @TurtlePU We can but we don't use is now since in this PR `<To/From>Polynomial` classes are ditched already.

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
iokasimov: No it's more readable since we don't use `fromPolynomial` anymore.

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
iokasimov: Those classes are already removed

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
TurtlePU: I mean, you can receive the same result by running `evalPolynomial` and picking `b` to be `Sources c i`

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
TurtlePU: I mean, this is just a map, I don't know if we need it as a separate function

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
TurtlePU: This can be done without unpacking: `evalMapPolynomial (\x -> if x == 0 then one else var x)`

iokasimov: @TurtlePU Thanks! I'll do it in next PR if you don't mine.d

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
TurtlePU: Of course, no prob!

src/ZkFold/Base/Algebra/Polynomials/Multivariate.hs
TurtlePU: No prob!

TurtlePU (approved):

PR #105, opened 2024-05-02
Improve univariate tests

iokasimov:
* Improve tests in `Univariate` module
* Define `evalPolynomial` with any element as a monomial
* Optimize expressions in `Multivariate` module
* Separate `Sources` into a module

Related issue: https://github.com/zkFold/zkfold-base/issues/22


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
```diff
@@ -144,46 +145,6 @@ circuits b = let (os, r) = runState b mempty in (\o -> r { acOutput = o }) <$> o
...
-
-instance Ord i => Semiring (Sources a i)
-
-instance Ord i => Ring (Sources a i)
-
-instance Ord i => Field (Sources a i) where
-    finv = id
-    rootOfUnity _ = Just (Sources mempty)
-
 instance (Finite a, Ord i) => BinaryExpansion (Sources a i) where
```
TurtlePU, line 148: What about this instance? Can it be moved to the `Sources` module, too?

TurtlePU (approved):

Issue #102, opened 2024-04-29
Refactor Symbolic & SymbolicData


echatav:
- [x] Move & modify `DiscreteField` from `Symbolic` to `Basic`
- [x] Create a `Trichotomy` class for discrete fields with a total ordering with law of trichotomy
- [x] Create `Trichotomy` instances for `Zp`
- [ ] Create `Trichotomy` instance for `ArithmeticCircuit`
- [x] Redefine `Symbolic` using `Trichotomy`
- [x] Create a `VectorSpace` class for vector spaces over a field with a `Basis` type
- [ ] Create lots of `VectorSpace` instances and relate to `Representable` functors and generics
- [x] Redefine `SymbolicData` using `VectorSpace`
- [ ] Reimplement all `SymbolicData` instances as `VectorSpace` and associated instances
- [x] Create a `Tensorial` class for multlinear maps of vector spaces with `InputBasis` and `OutputSpace` types
- [ ] Redefine `Arithmetizable` using new `SymbolicData`
- [ ] Reimplement all `Arithmetizable` instances
- [ ] Create comparison classes `Eq` and `Ord` for symbolic data with default Foldable implementations and lots of instances
- [x] Create algebraic classes (`AdditiveSemigroup`, etc) for `SymbolicData`
- [ ] Implement algebraic classes for `Identity`, `UInt n`
- [ ] Implement algebraic classes for `Vector n u`
- [ ] Replace list algorithms in UInt and ByteString with vector algorithms

echatav:
@TurtlePU 
Arithmetization looks like this then:
```
class Arithmetizable a f where
  type InputRep a f :: Type
  type OutputRep a f :: Type
  arithmetize :: f -> (InputRep a f -> ArithmeticCircuit a) -> OutputRep a f -> ArithmeticCircuit a

instance SymbolicData a u => Arithmetizable a (u (ArithmeticCircuit a)) where
  type InputRep a (u (ArithmeticCircuit a)) = Void
  type OutputRep a (u (ArithmeticCircuit a)) = Rep a u
  arithmetize c _ = index c

instance (SymbolicData a x, Arithmetizable a f) => Arithmetizable a (x (ArithmeticCircuit a) -> f) where
  type InputRep a (x (ArithmeticCircuit a) -> f) = Either (Rep a x) (InputRep a f)
  type OutputRep a (x (ArithmeticCircuit a) -> f) = OutputRep a f
[...]

PR #107, opened 2024-05-09
Algebra - Introduce 1 modified + 3 new classes

echatav:
This PR breaks down part of #102, refactoring `Symbolic` and `SymbolicData` by providing the algebraic classes that will be used. These classes are:

* `DiscreteField'`

This modifies an existing `DiscreteField` class which was higher level (Symbolic) and moves it lower level (Base). For now it has a ' mark on its name because the original class will be removed in a later PR. It will be used to refactor Symbolic Bool, and Eq.

* `TrichotomyField`

This moves the algebraic logic for ordering from Symbolic to Base. It will be used to refactor Symbolic Ord.

* `VectorSpace`

This class will be used to refactor `SymbolicData`. It includes a number of choices made after investigating different avenues.

* `FunctionSpace`

This class will be used to refactor `Arithmetizable`.


src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -545,3 +585,138 @@ instance FromConstant b a => FromConstant b (p -> a) where
...
+-- direct sum of vector spaces
+instance (VectorSpace a v, VectorSpace a u)
+  => VectorSpace a (v :*: u) where
+    type Basis a (v :*: u) = Either (Basis a v) (Basis a u)
+    tabulateV f = tabulateV (f . Left) :*: tabulateV (f . Right)
+    indexV (a :*: _) (Left  i) = indexV a i
+    indexV (_ :*: b) (Right j) = indexV b j
+
+-- tensor product of vector spaces
+instance (Representable u, VectorSpace a v)
```
TurtlePU, line 676: Why `Representable u` instead of `VectorSpace a u`?

src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -545,3 +585,138 @@ instance FromConstant b a => FromConstant b (p -> a) where
...
+
+--------------------------------------------------------------------------------
+
+{- | Class of vector spaces with a basis. More accurately, when @a@ is
+a `Field` then @v a@ is a vector space over it. If @a@ is a `Ring` then
+we have a free module, rather than a vector space. `VectorSpace` may also be thought of
+as a "monorepresentable" class, similar to `Representable` but with a fixed
+element type.
+-}
+class VectorSpace a v where
```
TurtlePU, line 597: :fire: :fire: :fire: 
There are so many definitions however, maybe we need a separate module for theory of vector spaces?

src/ZkFold/Base/Algebra/Basic/Class.hs
echatav: `VectorSpace a u` won't typecheck although `VectorSpace (v a) u` will but isn't what we want. The example tensor product will be useful for is when defining symbolic "container" types like `Vector` which _is_ compositions with a `Representable` so I think this suffices for our use case.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -93,6 +93,24 @@ instance Arithmetic a => BinaryExpansion (ArithmeticCircuit a) where
...
     fromBinary bits = circuit $ Haskell.traverse runCircuit bits >>= horner
 
+instance Arithmetic a => DiscreteField' (ArithmeticCircuit a) where
+    equal r1 r2 = isZeroC (r1 - r2)
+
+instance Arithmetic a => Trichotomy (ArithmeticCircuit a) where
+    trichotomy r1 r2 =
+        let
+            bits1 = binaryExpansion r1
+            bits2 = binaryExpansion r2
```
echatav, line 103: Are these least significant bit first or most significant bit first? I did the lexicographical fold assuming most significant bit first (I think).

src/ZkFold/Base/Algebra/Basic/Class.hs
echatav: 🤔 Now I'm thinking maybe `VectorSpace (v a) u` does make more sense. Since any `Representable u` can be thought of as `forall x. VectorSpace x u`, it's only a more specific constraint which will still work for `Vector`s. But, it has a "safer" implementation for its `indexV` since it won't call on the unsafe `index` function.

The `index` function for `Vector n` is unsafe since it just calls `V.Vector`'s `(!)` unsafe index function but its `indexV` function is safe: `indexV v i = fromMaybe zero (v !? i)`.

TurtlePU: In addition, this coarsely separates "collection" data types in Symbolic from "plain" data types: first ones are `Representable` while second are just `VectorSpace`

src/ZkFold/Base/Algebra/Basic/VectorSpace.hs
```diff
@@ -0,0 +1,147 @@
...
+
+The type @LinearMap a f => f@ should be equal to
+
+@(VectorSpace a v0, .. ,VectorSpace a vN) => vN a -> .. -> v1 a -> v0 a@
+
+which via uncurrying is equivalent to
+
+@(VectorSpace a v0, .. ,VectorSpace a vN) => (vN :*: .. :*: v1) a -> v0 a@
+-}
+class VectorSpace a (OutputSpace a f) => LinearMap a f where
```
TurtlePU, line 117: I'm not sure actually if symbolic functions are linear map, e.g. `\x y -> x // y` can be surely turned into a circuit, but is not linear in `y`

TurtlePU (approved):

PR #108, opened 2024-05-14
Sort monomials on constructing polynomials

iokasimov:
Since list literals is the only way we use to build polynomials, we sort out all monomials there.

Related issue: https://github.com/zkFold/zkfold-base/issues/14


vlasin:
1. `P` is still exposed in the whole package (as it can be exported from `ZkFold.Base.Algebra.Polynomials.Multivariate.Polynomial`). Could we remove this possibility?
2. CI reports warnings. Could you fix them, please?

vlasin (approved):

PR #115, opened 2024-05-22
FunctionSpace modifications

echatav:
Makes the definition of `FunctionSpace` more symmetric by replacing the `InputBasis` type family by the `InputSpace` type family. The input basis can be recovered from `Basis a (InputSpace a f)`. Also changes the methods `coindexV` and `cotabulateV` to `uncurryV` and `curryV` which directly conveys the connection of these functions with `InputSpace`. Also adds a new combinator `composeFunctions`.

Motivation: This is anticipatory to considerations about structured inputs for arithmetic circuits. We wouldn't want to parametrize circuit inputs by an `i :: Type` with `acInputs :: i -> Natural` because functions are not so tractable. We would rather parametrize circuit inputs by an `i :: Type -> Type` with `acInputs :: i Natural`. Just one example of tractability, we can determine the `length` or `maximum` of the latter if `i` is `Foldable`.

### DIFF
--- a/src/ZkFold/Base/Algebra/Basic/VectorSpace.hs
+++ b/src/ZkFold/Base/Algebra/Basic/VectorSpace.hs
@@ -152,3 +152,11 @@ instance {-# OVERLAPPING #-}
         pi2 (_ :*: v) = v
 
     curryV k x = curryV (k . (:*:) x)
+
+composeFunctions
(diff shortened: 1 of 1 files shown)


echatav:
Relates to #117

TurtlePU (approved):

PR #116, opened 2024-05-22
Trivial optimize

echatav:
Wave towards a solution for optimizing circuits


vlasin:
Should we add `optimize` to `compile` as well?

echatav:
> Should we add `optimize` to `compile` as well?

Well, I think we should change the types of both `solder` and `compile` as in #117 sketch. Right now it feels like the "true compile" function is inside `compileIO`.

vlasin (approved):

PR #119, opened 2024-05-24
Plonk test vector

iokasimov:
Related issue: https://github.com/zkFold/zkfold-base/issues/103

- [x]  Find a way to generate all data so the content doesn't depend on current implementation

I tested it this way:
```haskell
main = do
    test <- testVector @(Plonk 32 ByteString)
    let verified = test <&> \(s, i, p) -> verify @(Plonk 32 ByteString) s i p
    void $ traverse print verified
```


src/ZkFold/Base/Protocol/NonInteractiveProof.hs
```diff
@@ -81,3 +82,19 @@ proveAPI bsS bsW =
...
+
+testVector :: forall a .
+    NonInteractiveProof a =>
+    Arbitrary a =>
+    Arbitrary (Witness a) =>
+    Binary (Setup a) =>
+    Binary (Input a) =>
+    Binary (Proof a) =>
+    IO [(ByteString, ByteString, ByteString)]
+testVector = generate . vectorOf 10 $ (,)
```
vlasin, line 94: Please make the number of tests into a function argument.

vlasin (approved):

PR #121, opened 2024-05-29
Refactoring Symbolic et al.

echatav:
This PR does a major refactor of the Symbolic language.
* The `Symbolic` constraint is refactored to include `DiscreteField` and `TrichotomyField`
* The `SymbolicData` constraint is refactored to be based on `VectorSpace`
* The `Arithmetizable` constraint is replaced by `SymbolicFunction` based on `FunctionSpace`
* Symbolic `Bool`, `Eq` and `Ord` have been redesigned around `DiscreteField` and `TrichotomyField`
* Redefines Cardano datatypes as generic records

Questions
* Should symbolic `Eq` have a `SymbolicData` superclass?
* For now constraints are minimized to only what's needed, i.e. `FiniteField` or `Foldable`, etc. Should they constrained further to `Symbolic` and `SymbolicData`?

Left for follow-up PRs:
* replace list algorithms with `Data.Vector`
* remove old `SymbolicData` and `Arithmetizable` found in `ZkFold.Symbolic.Compiler.Arithmetizable`
* rethink `ULPC.Types`
* refactor `ArithmeticCircuit` with input & output `SymbolicData` parameters
* unify `EuclideanSemiring` and `EEA` and `BinaryExpansion` classes into a single `Euclidean` class
* Get rid of `Sig` constraint and only use `Symbolic` for `verifySignature`.


vlasin:
Question: how to convert `Par1 a` to `Uint 64 a`?

Also, those `Comp1` and `unComp1` look a bit ugly as well as the necessity to wrap `a` in `Par1 a` every time. Do we have plans to improve it in the future?

vlasin:
It seems we lost `Ord` for `UInt`. Could you bring it back?

src/ZkFold/Symbolic/Compiler.hs
```diff
@@ -30,18 +31,33 @@ import           ZkFold.Symbolic.Compiler.Arithmetizable
     8. ZkFold.Symbolic.Compiler
 -}
 
--- | Arithmetizes an argument by feeding an appropriate amount of inputs.
-solder :: forall a f . Arithmetizable a f => f -> [ArithmeticCircuit a]
-solder f = arithmetize f $ circuits $ replicateA (inputSize @a @f) input
+-- | Precompiles function `f` into a sequence of arithmetic circuits.
+precompile
+  :: forall a f . (Arithmetic a, FunctionSpace (ArithmeticCircuit a) f)
```
vks4git, line 36: How will `FunctionalSpace (ArithmeticCircuit a) f` will translate into ArithmeticCircuits with multiple outputs? It won't be a sequence of circuits anymore, just one circuit with several inputs

src/ZkFold/Symbolic/Algorithms/Hash/MiMC.hs
```diff
@@ -1,15 +1,17 @@
...
+                                                                 (^), (||))
 
 import           ZkFold.Base.Algebra.Basic.Class
-import           ZkFold.Symbolic.Types           (Symbolic)
+import           ZkFold.Symbolic.Algorithms.Hash.MiMC.Constants (mimcConstants)
 
 -- | MiMC-2n/n (Feistel) hash function.
 -- See https://eprint.iacr.org/2016/492.pdf, page 5
-mimcHash :: forall a . Symbolic a => [a] -> a -> a -> a -> a
+mimcHash :: Semiring a => [a] -> a -> a -> a -> a
```
vks4git, line 14: :+1:

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -47,8 +48,22 @@ import           ZkFold.Symbolic.Data.UInt
...
 --
 newtype ByteString (n :: Natural) a = ByteString [a]
-    deriving (Haskell.Show, Haskell.Eq, Generic, NFData)
-
+    deriving
+      ( Haskell.Show
+      , Haskell.Eq
+      , Generic
+      , NFData
+      , Haskell.Functor
```
vks4git, line 56: Are Functor, Foldable and Traversable used anywhere? 
With the approach I propose it won't be an instance of those anymore.
```
newtype ByteString (n :: Natural) (backend :: Natural -> Type -> Type) (a :: Type) = ByteString (backend n a)
    deriving (Haskell.Show, Haskell.Eq, Generic, NFData)
```

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -47,8 +48,22 @@ import           ZkFold.Symbolic.Data.UInt
...
+      , Haskell.Foldable
+      , Haskell.Traversable
+      )
+
+instance (FiniteField a, KnownNat n) => VectorSpace a (ByteString n) where
```
vks4git, line 61: Same context as above. With my approach, it won't work anymore. How bad is it?

src/ZkFold/Symbolic/Data/Ed25519.hs
```diff
@@ -1,115 +1,108 @@
...
-zpToEd (Point x y) = Point (toZp . toConstant $ x) (toZp . toConstant $ y)
-
-edToZp :: (Finite (Zp p)) => Point (Ed25519 Void) -> Point (Ed25519 (Zp p))
-edToZp Inf         = Inf
-edToZp (Point x y) = Point (fromConstant . fromZp $ x) (fromConstant . fromZp $ y)
+data PointEd25519 a = PointEd25519 (UInt 256 a) (UInt 256 a)
+  deriving stock (P.Eq, P.Functor, P.Foldable, P.Traversable, Generic1)
+deriving via Generically1 PointEd25519 instance FiniteField a => VectorSpace a PointEd25519
+instance (DiscreteField a, FiniteField a) => Eq a PointEd25519
+instance (TrichotomyField a, FiniteField a) => Ord a PointEd25519
```
vks4git, line 32: Wait, how does Ord work for two-dimensional points? Lexicografically?
Anyway, do we need this instance?

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -81,39 +97,43 @@ cast n =
...
+              ( (if Haskell.even iteration then b - oldS else oldS),
+                (if Haskell.odd iteration then a - oldT else oldT),
+                oldR )
+              (r Haskell.== zero)
+            where
+                quotient = oldR `div` r
+
+                recurse = eea' (iteration + 1) r (oldR - quotient * r) s (quotient * s + oldS) t (quotient * t + oldT)
+
+instance (Arithmetic a, KnownNat n) => EEA (UInt n (ArithmeticCircuit a)) where
```
vks4git, line 120: Why do we have to to implement eea twice now? Can't there be a unified implementation?

src/ZkFold/Symbolic/Data/Compare.hs
```diff
@@ -0,0 +1,82 @@
...
+lt = Ordering (negate one)
+eq = Ordering zero
+gt = Ordering one
+
+instance Ring a => Haskell.Semigroup (Ordering a) where
+  Ordering x <> Ordering y = Ordering (x * x * (x - y) + y)
+instance Ring a => Haskell.Monoid (Ordering a) where
+  mempty = eq
+
+class Eq a u => Ord a u where
```
vks4git, line 55: For context, see my comment about ByteString's proposed representation.
Do you see how can it fit in here? We'd like to see
```
compare :: ByteString n ArithmeticCircuit a -> ByteString n ArithmeticCircuit a -> Ordering (ArithmeticCircuit a)
```
Do you think making something like
```
newtype Ordering b a = Ordering (b a)
```
and thus separating field elements and computational backend will work? This is how I did it in my branch

src/ZkFold/Symbolic/Data/ByteString.hs
echatav: So long as `backend` has those instances then so will `ByteString`. Any symbolic data structure necessarily is a known length list of field elements so ought to be `Traversable` and `VectorSpace`.

src/ZkFold/Symbolic/Data/Ed25519.hs
echatav: Yes, lexicographically. Perhaps we don't need it.

src/ZkFold/Symbolic/Data/Compare.hs
echatav: Instead of using backends, if we have structured multi-outputs, then you can just directly use `ArithmeticCircuit (ByteString n) a` or `ByteString n (ArithmeticCircuit U1 a)`. I don't think backend is necessary.

src/ZkFold/Symbolic/Data/UInt.hs
echatav: Because I simplified `bool` to be only Symbolic, not Haskell's bool. It may be possible to generalize it back though I think the overall simpler design makes it worth it.

src/ZkFold/Symbolic/Compiler.hs
echatav: Once we have both structured inputs and structured multi-outputs, then the compile function will look something like this:

```Haskell
compile
  :: (Arithmetic a, SymbolicFunction (ArithmeticCircuit i Par1 a) f)
  => f -> ArithmeticCircuit (InputSpace f) (OutputSpace f) a
compile f = optimize (solder (curryV f (desolder idCircuit)))
```

echatav:
> It seems we lost `Ord` for `UInt`. Could you bring it back?

Not sure what you mean, we still have `Ord` for `UInt`s. I even added a test.

echatav:
> Question: how can we convert `Par1 a` to `Uint 64 a` and back?

I'm confused. These are different types, `Par1 a` is 1d and `UInt 64 a` is not.

> 
> Also, those `Comp1` and `unComp1` look a bit ugly as well as the necessity to wrap `a` in `Par1 a` every time. Do we have plans to improve it in the future? (I am looking at `batchTransfer` example)

I tried to make symbolic functions work for bare `a`s. It worked fine when the `a` was an input to a function, but when it's an output, then that caused irreconcilable overlapping instances because GHC cannot figure out if `(.... -> a)` means an output of `a` or a space of `a`.

vlasin:
> I'm confused. These are not isomorphic types, `Par1 a` is 1d and `UInt 64 a` is not.
Yes, but from time to time we will need to convert between the two in-circuit. It means that if the conversion is not possible the circuit should not be satisfiable, but if it does, then it performs the conversion. My question was, how can we do it now? Because `strictConv` only converts `a` to `Uint n a`.

> I tried to make symbolic functions work for bare `a`s. It worked fine when the `a` was an input to a function, but when it's an output, then that caused irreconcilable overlapping instances because GHC cannot figure out if `(.... -> a)` means an output of `a` or a space of `a`.

One of our end goals is to make Symbolic smart contract code look as clean as possible because this is what people will be doing - writing ZK smart contracts. Currently, we have two examples: `batchTransfer` and `randomOracle`. I get your point, though. Maybe we could introduce some helper functions to alleviate the effect of these changes? Could you take a look at these examples?

echatav:
> One of our end goals is to make Symbolic smart contract code look as clean as possible because this is what people will be doing - writing ZK smart contracts.

I appreciate this end goal. In my experience a clean interface naturally emerges from clean fundamentals. Clean fundamentals are what is motivating my designs and proposals. I try to follow the maxim of "model things as what they are".

src/ZkFold/Symbolic/Compiler.hs
echatav:
```Haskell
idCircuit :: SymbolicData i => Circuit i i a
idCircuit = mempty {acOutputs = iterateV 1 succ}
```

src/ZkFold/Symbolic/Data/Ed25519.hs
```diff
@@ -1,115 +1,108 @@
...
+    (fromConstant @Natural @(UInt 256 a) (15112221349535400772501151409588531511454012693041857206046113283949847762202))
+    (fromConstant @Natural @(UInt 256 a) (46316835694926478169428394003475163141307993866256225615783033603165251855960))
+
+isInf
+  :: (AdditiveMonoid (UInt 256 a), Eq a (UInt 256), Ring a)
+  => PointEd25519 a -> Bool a
+isInf (PointEd25519 x y) = x == zero && y == zero
+
+zpToEd :: (P.Eq a, AdditiveMonoid (UInt 256 a), ToConstant (UInt 256 a) P.Integer) => PointEd25519 a -> Base.Point (Base.Ed25519 Void)
+-- zpToEd pt | isInf pt        = Base.Inf
```
echatav, line 62: remove comment

src/ZkFold/Symbolic/Data/Compare.hs
echatav: Or perhaps you can instead define a type family like so:
```Haskell
type family Dim (a :: Type) (u :: Type -> Type) :: Natural
type instance Dim a (UInt n) = RegisterCount a n
```
Then you can use either `ArithmeticCircuit (Dim a u) a` or `u (ArithmeticCircuit 1 a)` as you need.

echatav: If you find a need to add methods to convert back and forth, then you can use a typeclass:

```Haskell
class HasDim a u where
  type family Dim a u :: Natural
  solder :: u (ArithmeticCircuit 1 a) -> ArithmeticCircuit (Dim a u) a
  desolder :: ArithmeticCircuit (Dim a u) a -> u (ArithmeticCircuit 1 a)
```

With structured outputs, `solder` and `desolder` can be implemented once without intermediate list (un)marshalling.

vks4git: Sorry, I still don't see how it all fits together. Can you elaborate further, please? Oh, and keep in mind that my PR focuses ONLY on performance. I totally agree that your approach is beautiful, I just want Symbolic to stop doing redundant computations. 

If we have `ByteString n (ArithmeticCircuit Par1 a)`, then what is the definition of ByteString itself? Does it hold a list (or any other container) of ArithmeticCircuits? If so, then evaluating ByteStrings will involve doing the same computations over and over again because all arithmetic circuits will share the same variables and constraints. 
If we have `ArithmeticCircuit (ByteString n) a`, how do we implement ByteStrings on Zp? Will it be a completely unrelated type?
The implementation of ByteStrings and UInts proposed by me solves both problems. ByteString n ArithmeticCircuit a is just a newtype over ArithmeticCircuit n a (a single circuit with n outputs, which somewhat resembles your second option). ByteString on Zp's is just ByteString n Vector (Zp p) which is just a newtype over Vector n (Zp p).
So we have the same newtype being able to hold both ArithmeticCircuits without any redundand computations and Zp's.

src/ZkFold/Symbolic/Data/ByteString.hs
vks4git: But ArithmeticCircuit is not a Functor because of witness

src/ZkFold/Symbolic/Data/Compare.hs
echatav: > If we have ByteString n (ArithmeticCircuit Par1 a), then what is the definition of ByteString itself?

The definition of `ByteString n` is the same as it is now on the main branch:

```Haskell
newtype ByteString (n :: Natural) a = ByteString [a]
```

> Does it hold a list (or any other container) of ArithmeticCircuits?

`ByteString n` is a list newtype so it can hold a list of circuits or of anything.
`ByteString n (ArithmeticCircuit Par1 a)` holds a list of single-output circuits.

> If so, then evaluating ByteStrings will involve doing the same computations over and over again because all arithmetic circuits will share the same variables and constraints.

That's true, which is why you want to compute things on a multi-output, single-circuit `ArithmeticCircuit (ByteString n) a` instead.

> how do we implement ByteStrings on Zp?

You can directly use `ByteString n (Zp p)` as in the main branch.

> Will it be a completely unrelated type?

It is the exact same symbolic datatype `ByteString n :: Type -> Type`.

> ByteString n ArithmeticCircuit a is just a newtype over ArithmeticCircuit n a (a single circuit with n outputs, which somewhat resembles your second option).

It is equivalent to my second option, which is a single circuit with `n` outputs, `ArithmeticCircuit (ByteString n) a`.

With `UInt`, the second option is a single circuit with `numberOfRegisters a n` outputs, `ArithmeticCircuit (UInt n) a`.

Instead of using a heterogeneous vector type `Vector dim` to contain the outputs of the circuit, which forces you to calculate `dim` at the type level from `n` and `a`, you use homogeneous symbolic datatypes (vector spaces) `ByteString n` or `UInt n` to achieve the same effect.

your proposal is equivalent to my proposal as:
* your `u Vector a` is equivalent to my (and main branch's) `u a`
* your `u ArithmeticCircuit a` is equivalent to my `ArithmeticCircuit u a`
* your multi-output single-circuit `KnownNat n => ArithmeticCircuit n a` is equivalent to my `(VectorSpace a o, Traversable o) => ArithmeticCircuit o a`

echatav:
Will have to rethink how to add this now with recent merges. Probably need to start with refactoring circuit outputs before this is feasible again.

PR #123, opened 2024-06-03
WIP multi-output circuits

vks4git:


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
```diff
@@ -69,7 +71,7 @@ type ClosedPoly i a = forall x . Algebra a x => (i -> x) -> x
 -- NOTE: the property above is correct by construction for each function of a
 -- suitable type, you don't have to check it yourself.
 
-class Monad m => MonadBlueprint i a m | m -> i, m -> a where
+class Monad m => MonadBlueprint i a n m | m -> i, m -> a, m -> n where
```
TurtlePU, line 74: With separate `Circuit` type in place, we can work with circuits of arbitrary size and not have `n`  as a class parameter

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -51,31 +65,94 @@ data ArithmeticCircuit a = ArithmeticCircuit
...
+               acVarOrder = empty,
+               acRNG      = mkStdGen 0
+           }
+
+instance Eq a => Semigroup (ArithmeticCircuit n a) where
+    r1 <> r2 =
+        ArithmeticCircuit
+            {
+                acCircuit = acCircuit r1 <> acCircuit r2
+            ,   acOutput  = Z.zipWith max (acOutput r1) (acOutput r2)
```
vlasin, line 131: Previously, we prioritized the left circuit here. We should carefully check that it does not introduce new errors.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -28,75 +38,85 @@ import           ZkFold.Symbolic.Data.Eq
...
     (^) = natPow
 
-instance Arithmetic a => MultiplicativeMonoid (ArithmeticCircuit a) where
+instance (Arithmetic a, KnownNat n) => MultiplicativeMonoid (ArithmeticCircuit n a) where
     one = mempty
 
-instance (Arithmetic a, FromConstant b a) => FromConstant b (ArithmeticCircuit a) where
-    fromConstant c = embed (fromConstant c)
+-- TODO: The constant will be replicated in all outputs. Is this the desired behaviour?
+instance (Arithmetic a, FromConstant b a, KnownNat n) => FromConstant b (ArithmeticCircuit n a) where
```
vlasin, line 90: Do we want this instance? Or just the one for `ArithmeticCircuit 1 a`?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -28,75 +38,85 @@ import           ZkFold.Symbolic.Data.Eq
...
+instance (Arithmetic a, KnownNat n) => Field (ArithmeticCircuit n a) where
     finv = invertC
-    rootOfUnity n = embed <$> rootOfUnity n
-
-instance Arithmetic a => BinaryExpansion (ArithmeticCircuit a) where
-    binaryExpansion r = circuits $ runCircuit r >>= expansion (numberOfBits @a)
-    fromBinary bits = circuit $ Haskell.traverse runCircuit bits >>= horner
-
-instance Arithmetic a => DiscreteField' (ArithmeticCircuit a) where
+    rootOfUnity n = embedAll <$> rootOfUnity n
```
vlasin, line 102: This should be just `embed`, right?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -28,75 +38,85 @@ import           ZkFold.Symbolic.Data.Eq
...
+
+-- TODO: The only implementation that seems to make sense is when there is only one output.
+-- It is true?
+--
+-- Anyway, @binaryExpansion@ of an arithmetic circuit will return copies of the same circuit with different outputs.
+-- The whole point of this refactor was to avoid this.
+--
+-- Ideally, we want to return another ArithmeticCircuit with a number of outputs corresponding to the number of bits.
+-- This does not align well with the type of @binaryExpansion@
```
vlasin, line 111: I agree with this comment. Is it possible to change the signature of the `binaryExpansion` function? Maybe introduce a type family to handle both cases?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -28,75 +38,85 @@ import           ZkFold.Symbolic.Data.Eq
...
-
-instance Arithmetic a => BinaryExpansion (ArithmeticCircuit a) where
-    binaryExpansion r = circuits $ runCircuit r >>= expansion (numberOfBits @a)
-    fromBinary bits = circuit $ Haskell.traverse runCircuit bits >>= horner
-
-instance Arithmetic a => DiscreteField' (ArithmeticCircuit a) where
+    rootOfUnity n = embedAll <$> rootOfUnity n
+
+-- TODO: The only implementation that seems to make sense is when there is only one output.
+-- It is true?
```
vlasin, line 105: Yes, probably. It would be too unintuitive to consider expansions of multiple numbers at the same time.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
TurtlePU: We need this as `Semiring x` requires `FromConstant Natural x` and `Ring x` requires `FromConstant Integer x`

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
TurtlePU: +1, we also might want to have an `instance BinaryExpansion (UInt n a)`, but the current interface would force us to waste multitudes of registers

I propose we will redesign a `class BinaryExpansion` in one of our next PRs

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
TurtlePU: Well, given the laws for `BinaryExpansion`, an `instance Semiring (ArithmeticCircuit n a)` hints that there is one instance where in `binaryExpansion r :: [ArithmeticCircuit n a]` the `i`-th output of a `j`-th circuit is a `j`-th bit of an `i`-th output of `r`.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -28,75 +38,85 @@ import           ZkFold.Symbolic.Data.Eq
...
-        j <- runCircuit r2
-        newAssigned (\x -> x i * x j)
-
-instance Arithmetic a => Exponent (ArithmeticCircuit a) Natural where
+instance (Arithmetic a, KnownNat n) => SymbolicData a n (ArithmeticCircuit n a) where
+    pieces = id
+
+    restore = ArithmeticCircuit
+
+-- TODO: I had to add these constraints and I don't like them
```
TurtlePU, line 46: This happens because in `Finite`, I specified that `Order a` and `NumberOfBits a` have to be `KnownNat`. Maybe there is a better approach?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
```diff
@@ -3,45 +3,60 @@
...
-boolCheckC r = circuit $ do
-    i <- runCircuit r
-    newAssigned (\x -> x i * (x i - one))
+boolCheckC r = circuitN $ do
+    is <- runCircuit r
+    for is $ \i -> newAssigned (\x -> x i * (x i - one))
 
-embed :: Arithmetic a => a -> ArithmeticCircuit a
+
+-- | TODO: Think about circuits with multiple outputs
```
TurtlePU, line 40: Maybe `embed :: Vector n a -> ArithmeticCircuit n a`?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
TurtlePU: BTW, I guess we can introduce the following instances:
```haskell
instance (FromConstant b a, KnownNat n) => FromConstant b (Vector n a) -- by replicate
instance (FromConstant b (Vector n a)) => FromConstant b (ArithmeticCircuit n a) -- by embed
```

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: If we replace the type of the state in `MonadBlueprint` instance with `Circuit`, we will not need `Monoid (ArithmeticCircuit n a)` instances at all, only the `Monoid (Circuit a)` ones

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
```diff
@@ -90,10 +92,10 @@ class Monad m => MonadBlueprint i a m | m -> i, m -> a where
     input :: m i
 
     -- | Returns a circuit with supplied variable as output.
-    output :: i -> m (ArithmeticCircuit a)
+    output :: Vector n i -> m (ArithmeticCircuit n a)
```
TurtlePU, line 95: I feel like now we do not use this method anywhere and we better remove it so we can have pure instances. Maybe in one of the next PRs

src/ZkFold/Symbolic/Compiler/Arithmetizable.hs
```diff
@@ -15,102 +16,98 @@ import           Numeric.Natural                                     (Natural)
...
+    pieces :: x -> ArithmeticCircuit n a
 
-    -- | Restores `x` from the circuits' outputs.
-    restore :: [ArithmeticCircuit a] -> x
+    -- | Restores `x` from the circuit's outputs.
+    restore :: Circuit a -> Vector n Natural -> x
 
     -- | Returns the number of finite field elements needed to describe `x`.
     typeSize :: Natural
+    typeSize = value @n
```
TurtlePU, line 37: Any other implementation would not make sense, so we better make this a separate function instead of a class member

src/ZkFold/Symbolic/Compiler/Arithmetizable.hs
```diff
@@ -15,102 +16,98 @@ import           Numeric.Natural                                     (Natural)
...
-    arithmetize :: x -> [ArithmeticCircuit a] -> [ArithmeticCircuit a]
+    arithmetize :: x -> ArithmeticCircuit i a -> ArithmeticCircuit o a
 
     -- | The number of finite field elements needed to describe an input of `x`.
     inputSize :: Natural
+    inputSize = value @i
 
     -- | The number of finite field elements needed to describe the result of `x`.
     outputSize :: Natural
+    outputSize = value @o
```
TurtlePU, line 93: Same comment as for `typeSize`

src/ZkFold/Symbolic/Data/Combinators.hs
```diff
@@ -83,6 +89,49 @@ highRegisterSize = getNatural @n -! registerSize @a @n * (numberOfRegisters @a @
 registerSize :: forall a n . (Finite a, KnownNat n) => Natural
 registerSize = Haskell.ceiling (getNatural @n % numberOfRegisters @a @n)
 
+
+--type family RegisterSize (a :: k) (bits :: Natural) :: Natural where
+--    RegisterSize a bits =
+--        OrdCond (CmpNat bits ((Div bits (NumberOfRegisters a bits)))
+
+type family NumberOfRegisters (a :: Type) (bits :: Natural) :: Natural where
+    NumberOfRegisters a bits = NumberOfRegisters' a bits (ListRange 1 1000) -- TODO: Compilation takes ages if this constant is greater than 10000.
```
TurtlePU, line 98: I agree, this bound makes sense

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -48,13 +49,20 @@ import           ZkFold.Symbolic.Data.Eq.Structural
...
 -- TODO (Issue #18): hide this constructor
-data UInt (n :: Natural) a = UInt ![a] !a
+data family UInt :: Natural -> Type -> Type
+
+data instance UInt (n :: Natural) (Zp p) = UIntZp ![Zp p] !(Zp p)
     deriving (Haskell.Show, Haskell.Eq, Generic, NFData)
 
-instance (FromConstant Natural a, Finite a, AdditiveMonoid a, KnownNat n) => FromConstant Natural (UInt n a) where
-    fromConstant = Haskell.fst . cast @a @n . (`Haskell.mod` (2 ^ getNatural @n))
+data instance UInt (n :: Natural) (ArithmeticCircuit n a) = UIntAc (ArithmeticCircuit (NumberOfRegisters a n) a)
```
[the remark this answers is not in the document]
TurtlePU: Also maybe the `Circuit a` would be a better fit now instead of `ArithmeticCircuit` as it is not indexed

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
vlasin: Yes, on second thought, it all makes sense and might be useful. For example, if we need to add the same constant to a bunch of variables at once.

echatav: Having the algebraic instances for n/=1 feels weird to me

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
echatav: Monoid instance and circuit in state/blueprint should have outputs n=0

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/MonadBlueprint.hs
echatav: You don't _need_ a separate type but `n` should equal `0` in the state updates. Outputs should be passed in the return variable, not the state variable.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
echatav: I think we had better look at the underlying algebra a little more carefully. For instance I believe that `BinaryExpansion` is really an internal encoding of the Euclidean degree function of a Euclidean semiring. 

> Ideally, we want to return another ArithmeticCircuit with a number of outputs corresponding to the number of bits.

I don't think this is possible is it? The number of bits isn't static; i.e. the binary expansion of `0` is `[]` but the binary expansion of `1` is `[1]`, etc.

echatav:
My belief is that the outputs should be parameterized by a `o :: Type -> Type` rather than `n :: Natural` and
* `Semigroup/Monoid` instances should be on `ArithmeticCircuit U1 a`
* state/bluprint updates should have a state variable of `ArithmeticCircuit U1 a`
* algebraic instances, particularly multiplication instances, should be on `ArithmeticCircuit Identity a`

src/ZkFold/Symbolic/Data/UInt.hs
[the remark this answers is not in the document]
echatav: ^ With `Type -> Type` parametrized outputs, we should have two forms in general for algebraic data `u` or specifically for `UInt n` to consider, like so:
```
circuit :: ArithmeticCircuit (UInt n) a
circuits :: UInt n (ArithmeticCircuit Identity a)
```
Or generally,
```
circuit :: ArithmeticCircuit o a
circuits :: o (ArithmeticCircuit Identity a)
```
There are natural ways to translate between them when `o` is `Functor` and `Foldable`.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
echatav: If we have `Type -> Type` parametrized outputs, then as an intermediate step you can have the binary expansion be a circuit `binaryExpansionC :: ArithmeticCircuit Identity a -> ArithmeticCircuit [] a`

TurtlePU: > I don't think this is possible if outputs are statically sized is it? 
In case of arithmetic circuits, the length of a list returned is always the same and is equal to `numberOfBits @a`

src/ZkFold/Symbolic/Data/UInt.hs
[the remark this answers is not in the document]
echatav: > Forcing n to be the same in UInt and in ArithmeticCircuit in LHS seems odd to me, do they have to be related?

I agree, I think in the circuit should be `n=1` then like so:

```
data instance UInt (n :: Natural) (ArithmeticCircuit 1 a) = UIntAc (ArithmeticCircuit (NumberOfRegisters a n) a)
```

echatav:
> Maybe `U0`?

`U1` is the 0-dimensional container type, `data U1 x = U1`, the `U` is for unit and `1` is for number of kind arguments.

> Maybe on U1, too?

The algebraic instances as they currently exist make a lot of sense for single (1-dimensional) output because a single-output circuit represents a function `y = f(x1,..,xi)` and given two such functions, we have a natural product, y1*y2 = f1(x1,..,xi)*f2(x1,..,xi). The product for multi-output or zero-output functions is definable, just not something we currently support, and not very natural or standard.

> This would make smart contracts more readable

If there's a clear use case I'd support it and I'd like to see the details. Perhaps though, that's something for the future and we should limit scope for brand new interfaces.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
echatav: Ok, that makes sense and I just tested it a little in GHCi.

echatav: I still don't think that `Vector (NumberOfBits a)` is _needed_ but either it or `[]` can be used as wanted.

echatav:
For context for some of my review comments: while working on symbolic/arithmetizable, I also initially followed the siren call of adding a type level `Natural`, that is promoting `typeSize :: SymbolicData a u => Natural` which led down a similar road to this, including the need to promote register count computations. But if the only real use of that natural is in a `Vector n`, then we can avail ourselves of very powerful typeclass hierarchies if you replace `Natural` with `Type -> Type` and `KnownNatural` with some typeclass that represents "containers of known length". A known length means it's known at runtime and so there ends up being no need for type level calculations. The class for known length container, regardless of what it contains is [`Representable`](https://hackage.haskell.org/package/adjunctions-4.4.2/docs/Data-Functor-Rep.html#t:Representable). That doesn't quite work for symbolic data such as `UInt n` because as you know the size of a `UInt n a` depends on the size of the finite field `a` _and_ that dependency is replicated in its representation as a circuit. So, the class we need for that is what I called `VectorSpace`. Both `VectorSpace` and `Representable` sort of "intersect" with stock derived `Functor`/`Foldable`/`Traversable` in a compatible way.

src/ZkFold/Symbolic/Data/UInt.hs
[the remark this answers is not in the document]
TurtlePU: > ^ With `Type -> Type` parametrized outputs, we should have two circuit forms in general for symbolic data `u` or specifically for `UInt n` to consider

That's a really wonderful observation! This way we can convert between "packed" and "unpacked" representations of data types, switching between efficiency and ergononics where applicable. 

But I guess we should stick to the more performant version of `ArithmeticCircuit (UInt n) a` whenever possible

TurtlePU:
@echatav, my bad, messed up the names of types in https://github.com/zkFold/zkfold-base/pull/123#issuecomment-2150351702, sorry x) 

I completely agree that algebraic instances should be written out for 1-output circuits only, it's just that the name of the corresponding functor, `Identity`, is too long and recalls the association with monads and functor algebras, while this is most likely going to be used as some kind of "base type" in computations, so maybe some other name would be more suitable

echatav:
@TurtlePU Would you prefer we used `Par1`? It's in base, has all instances I'd expect and has a shorter name.

TurtlePU:
@echatav, yes, good one!

echatav:
```Haskell
type family Vars (n :: Natural) :: Type -> Type where
  Vars 0 = U1
  Vars n = Par1 :*: Vars (n-1)
```
^_^

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -51,31 +65,94 @@ data ArithmeticCircuit a = ArithmeticCircuit
...
+        -- ^ The witness generation function
+        acVarOrder :: Map (Natural, Natural) Natural,
+        -- ^ The order of variable assignments
+        acRNG      :: StdGen
+    } deriving (Generic, NFData)
+
+data ArithmeticCircuit n a = ArithmeticCircuit
+  { acCircuit :: Circuit a
+  , acOutput  :: Vector n Natural
+  } deriving (Generic, NFData)
```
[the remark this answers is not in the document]
vks4git: This PR focuses on performance. I separated circuits and outputs so that when evaluating circuits, we can supply one circuit and only change output variables. See instance SymbolicData a (Vector n a), for example

src/ZkFold/Base/Algebra/EllipticCurve/Ed25519.hs
```diff
@@ -14,9 +14,9 @@ import           ZkFold.Base.Algebra.Basic.Number
...
 
 -- | The Ed25519 curve used in EdDSA signature scheme.
--- @a@ represents the "computational backend" used to store and perform operations on curve points.
+-- @b@ represents the "computational backend" used to store and perform operations on curve points using elements of @a@.
 --
-data Ed25519 a
+data Ed25519 b a
 
 -- | 2^252 + 27742317777372353535851937790883648493 is the order of the multiplicative group in Ed25519
 -- with the generator point defined below in @instance EllipticCurve (Ed25519 Void)@
```
vlasin, line 22: @instance EllipticCurve (Ed25519 Void Void)@

TurtlePU (approved):

PR #124, opened 2024-06-03
New setting for the blockchain network

zlonast:


src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -38,14 +37,14 @@ type G2 = Point BLS12_381_G2
     NOTE: we need to parametrize the type of transcripts because we use BuiltinByteString on-chain and ByteString off-chain.
     Additionally, we don't want this library to depend on Cardano libraries.
 -}
-data Plonk (d :: Natural) t = Plonk F F F (Map Natural F) (ArithmeticCircuit F) F
```
vlasin, line 41: Here we should have `[Natural]` instead of a `Map`

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -58,9 +57,10 @@ data PlonkSetupParams = PlonkSetupParams {
         omega :: F,
         k1    :: F,
         k2    :: F,
-        gs    :: V.Vector G1,
```
vlasin, line 61: We need `gs` in the setup for the prover, though we do not need this field for the verifier. Consider Splitting the type for better `verify` efficiency.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -117,13 +117,52 @@ instance Arbitrary PlonkProverSecret where
...
 
+instance Arbitrary PlonkInput where
+    arbitrary = do
+        PlonkInput . negate <$> arbitrary
+
 data PlonkProof = PlonkProof G1 G1 G1 G1 G1 G1 G1 G1 G1 F F F F F F
     deriving (Show)
 
--- TODO (Issue #18): make the code safer, check list lengths (?)
+commonSetup :: forall d .
```
vlasin, line 130: We do not need this function since the setup should be run only once per circuit. But the `prove` algorithm can be run many times with different inputs.

src/ZkFold/Base/Protocol/ARK/Plonk/Internal.hs
```diff
@@ -92,20 +92,19 @@ fromPlonkConstraint (ql, qr, qo, qm, qc, a, b, c) =
...
 
-addPublicInput :: Natural -> F -> [Polynomial' F] -> [Polynomial' F]
-addPublicInput i _ ps = var i : ps
+addPublicInput :: Natural -> [Polynomial' F] -> [Polynomial' F]
+addPublicInput i ps = var i : ps
 
 removeConstantVariable :: (Eq c, Field c, Scale c c, FromConstant c c) => Polynomial' c -> Polynomial' c
 removeConstantVariable = evalPolynomial evalMapM (\x -> if x == 0 then one else var x)
 
-toPlonkArithmetization :: forall a . KnownNat a => Map Natural F -> ArithmeticCircuit F
```
vlasin, line 101: This should also take a `Vector` as the first argument.

src/ZkFold/Base/Protocol/Commitment/KZG.hs
```diff
@@ -56,8 +56,8 @@ instance forall (c1 :: Type) (c2 :: Type) t f d kzg .
     type Transcript (KZG c1 c2 t f d)   = ByteString
     type Setup (KZG c1 c2 t f d)        = (V.Vector (Point c1), Point c2, Point c2)
     type Witness (KZG c1 c2 t f d)      = WitnessKZG c1 c2 t f d
-    type Input (KZG c1 c2 t f d)        = Map f (V.Vector (Point c1), V.Vector f)
-    type Proof (KZG c1 c2 t f d)        = Map f (Point c1)
+    type Input (KZG c1 c2 t f d)        = ()
+    type Proof (KZG c1 c2 t f d)        = (Map f (V.Vector (Point c1), V.Vector f), Map f (Point c1))
```
vlasin, line 60: What is the reason for this change? Please revert.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -38,14 +39,14 @@ type G2 = Point BLS12_381_G2
     NOTE: we need to parametrize the type of transcripts because we use BuiltinByteString on-chain and ByteString off-chain.
     Additionally, we don't want this library to depend on Cardano libraries.
 -}
-data Plonk (d :: Natural) t = Plonk F F F (Map Natural F) (ArithmeticCircuit F) F
+data Plonk (d :: Natural) t = Plonk F F F Natural (V.Vector Natural) (ArithmeticCircuit F) F
```
vlasin, line 42: Can we use `ZkFold.Base.Data.Vector.Vector` here? It has the size type parameter in it.

vlasin (approved):

PR #128, opened 2024-06-05
feat: use `foldl'` instead of `foldl` in `length` prelude function

sourabhxyz:
Just started studying the codebase of this awesome project :D. Not sure if there's a good reason of keeping `foldl` here 🤔.


echatav:
Good catch, `foldl'` matches the default implementation of `Integer` `length`. Not sure why CI is failing but it's not to do with the code I surmise.

vlasin (approved):

PR #131, opened 2024-06-07
fix: correct `Binary` instance of `LittleEndian` type

sourabhxyz:
Related to https://github.com/zkFold/zkfold-base/issues/130.

Not sure if current behaviour is as expected, but this is what I had in mind.

"Binary instance" test pass by this modification. Use of `reverse` can be avoided by modifying `accum` function accordingly.


echatav:
> Use of reverse can be avoided by modifying accum function accordingly.
Can you make this change?

sourabhxyz:
Hi @echatav, I have made the required change, please have a look.

echatav (approved):
Thanks!

PR #134, opened 2024-06-11
Define Applicative/Maybe instances for Symbolic Maybe

iokasimov:
In order to use `Symbolic` `Maybe` in `P2P` example as a regular `Haskell` one, I've tried to define these instances.

The interesting part here is that I force these `Maybe` values to have either `zero` or `one` status. Not sure if it's correct to do in circuits.

### DIFF
--- a/src/ZkFold/Symbolic/Data/Maybe.hs
+++ b/src/ZkFold/Symbolic/Data/Maybe.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE DerivingStrategies    #-}
-{-# LANGUAGE UndecidableInstances  #-}
 {-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
 module ZkFold.Symbolic.Data.Maybe (
     Maybe, maybe, just, nothing, fromMaybe, isNothing, isJust, find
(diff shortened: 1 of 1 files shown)


echatav:
Generally speaking, symbolic data won't have `Applicative` instances. They will however have `VectorSpace` instances which has related functions:

```Haskell
pureV :: VectorSpace a v => a -> v a
zipWithV :: VectorSpace a v => (a -> a -> a) -> v a -> v a -> v a
```

They definitely won't be `Monad`s or anything like that. That is, a `Monad u => Monad (Maybe u)` won't help you because you don't have a `Monad (UInt n)` instance.

[Here](https://github.com/zkFold/zkfold-base/blob/dfa90287cbc570e13bda2600be879f6924d0362a/src/ZkFold/Symbolic/Data/Maybe.hs) is how it will look when symbolic data is refactored.

iokasimov:
@echatav Okay, so it looks like we need to wait a little bit.

echatav:
Hmmm, it should be possible to work backwards from the `VectorSpace` implementation of `bindMaybe` to a `SymbolicData` implementation. Probably its type signature should like like:

```Haskell
bindMaybe
  :: SymbolicData a v
  => (u (ArithmeticCircuit a) -> Maybe v (ArithmeticCircuit a))
  -> Maybe u (ArithmeticCircuit a) -> Maybe v (ArithmeticCircuit a)
```

If you define `(>>=) = flip bindMaybe` and `return = just`, then you can use `QualifiedDo` or `RebindableSyntax` to get `do` notation on symbolic `Maybe`.

PR #139, opened 2024-06-18
Draft for wrapper & Blake2b

zlonast:
We want to write a wrapper for a symbolic smart contract that would wrap around a circuit and accept one public input and many private ones. `ContractWrapper`
To do this, we need to be able to make logins [public or private.](https://github.com/zkFold/zkfold-base/pull/156)
To create a transaction hash, we must turn it into a Data with [toBuiltinData](https://github.com/zkFold/zkfold-base/pull/146), serialize it with [serialiseData](https://github.com/zkFold/zkfold-base/pull/147), and hash it with `blake2b_224`.

Reference impl [blake2b](https://www.rfc-editor.org/rfc/rfc7693.txt)

- [x] ContractWrapper 
- [x] Blake2b


src/ZkFold/Symbolic/Cardano/Contracts/SymbolicWrapper.hs
```diff
@@ -0,0 +1,98 @@
...
+
+type HashFun inputs rinputs outputs tokens datum a =
+    ( FromConstant (ByteString 224 a) (Public F))
+
+-- | Use wrapper to convert circuit to plonk circuit
+symbolicWrapper :: forall a privates inputs rinputs outputs tokens datum .
+    ( Eq (Bool a) (Public F),
+      Iso (HashData inputs rinputs outputs tokens datum a) a,
+      HashFun inputs rinputs outputs tokens datum a) =>
+    HashData inputs rinputs outputs tokens datum a -> Contract (privates + 1) a -> Public F -> Privates privates a -> Bool a
```
vlasin, line 90: This function's signature is off. 

- [ ] `Contract` should be the first argument because we will call `compile` on `symbolicWrapper contract`.
- [ ] Instead of `Public F`, I believe it should be `Public a`.

src/ZkFold/Symbolic/Cardano/Contracts/SymbolicWrapper.hs
```diff
@@ -0,0 +1,98 @@
...
+    ( KnownNat keylen
+    , keylen <= 64
+    , ToWords (ByteString (keylen * 8) a) (ByteString 8 a)
+    , Concat (ByteString 8 a) (ByteString 224 a)
+    ) => ByteString (keylen * 8) a -> ByteString 224 a
+blake2b_224 = blake2b_libsodium @keylen @28
+-}
+
+hashFunction :: HashData inputs rinputs outputs tokens datum a -> Public F
+hashFunction = undefined -- toConstant . fromConstant . blake2b_224 . serialiseData . toBuiltinData
```
vlasin, line 60: This function should return `Public a`.

src/ZkFold/Symbolic/Cardano/Contracts/SymbolicWrapper.hs
```diff
@@ -0,0 +1,98 @@
+{-# LANGUAGE TypeOperators #-}
+module ZkFold.Symbolic.Cardano.Contracts.SymbolicWrapper where
```
vlasin, line 2: I would rename this module like this: `ZkFold.Symbolic.Cardano.ContractWrapper`.

vlasin:
In this PR, we included

- [ ] a basic implementation of Blake2b in Symbolic
- [ ] a smart contract wrapper prototype for Cardano

What is left for the next PRs

- [ ] a proper testing for both numeric and symbolic Blake2b
- [ ] optimization of Blake2b for circuit compilation (i.e., replace uint operations with bit operations, use `ST` monad?)
- [ ] implementation of transaction data hashing which is consistent with the corresponding Plutus computations

vks4git (approved):

src/ZkFold/Base/Data/Vector.hs
```diff
@@ -133,6 +134,11 @@ unsafeConcat = Vector . concatMap fromVector
 chunks :: forall m n a . KnownNat n => Vector (m * n) a -> Vector m (Vector n a)
 chunks (Vector lists) = Vector (Vector <$> chunksOf (fromIntegral $ value @n) lists)
 
+instance IsList (Vector size a) where
+    type Item (Vector size a) = a
+    fromList = unsafeToVector
```
TurtlePU, line 139: Do we actually want to call an "unsafe" function implicitly? Maybe use the safe alternative to cast from lists

vlasin: Removed

PR #141, opened 2024-06-22
Refactor Symbolic Cardano types

vlasin:
- [x] Refactored `lockedByTxId` example
- [x] Put Cardano-related types into separate modules for better readability
- [x] Add missing methods
- [x] Add `txMint` field to the `Transaction` structure (probably temporary as it is redundant)
- [x] Add Random Oracle smart contract example


src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
```diff
@@ -0,0 +1,69 @@
...
+
+randomOracle :: forall a' a . (Symbolic a, Sig a, FromConstant a' a) => a' -> Tx a -> a -> Bool a
+randomOracle c tx w =
+    let -- The secret key is correct
+        condition1 = fromConstant @a' @a c == mimcHash mimcConstants zero w zero
+
+        -- Extracting information about the transaction
+        seed              = hash @a $ txiOutputRef $ head (fromVector $ txInputs tx)
+        Value (Vector xs) = txoTokens $ head $ fromVector $ txOutputs tx
+        (p, (name, n))    = xs !! 1
```
zlonast, line 54: Don't we want a list of one element here? I think about `xs !! 0`

src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
```diff
@@ -0,0 +1,69 @@
...
+        policyId          = fst $ head $ fromVector $ getValue $ txMint tx
```
zlonast, line 55: I think we should add `headV = head . fromVector` to `ZkFold.Base.Data.Vector`

src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
zlonast: or `Value (Vector [(p, (name, n))]) = txoTokens $ head $ fromVector $ txOutputs tx`

src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
```diff
@@ -0,0 +1,69 @@
...
+        -- The token's name is correct
+        condition2 = name == ByteString (binaryExpansion r)
+
+        -- The token's policy is correct
+        condition3 = p == policyId
+
+        -- The token's quantity is correct
+        condition4 = n == one
+
+    in condition1 && condition2 && condition3 && condition4
```
zlonast, line 69: Perhaps conditions should have names?
`conditionSecretKey && conditionTokenName && conditionTokenPolicy && conditionTokenQuantity`

zlonast (commented):
~~I think `xs !! 1` is a bug, but overal~~ I think it's a good changes.

src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
vlasin: The first element is ada because it has an empty bytestring as a policy ID. We expect the second element to be the token that we are minting.

src/ZkFold/Symbolic/Cardano/Contracts/RandomOracle.hs
vlasin: I think a component selector would be better.

vks4git:
Looks like it does not compile with -Werror, is it so?

vlasin:
> Looks like it does not compile with -Werror, is it so?

It is the same error as in `main`. We need to fix it and enable `Pedantic` in CI.

PR #145, opened 2024-06-24
arbitrary instance for Plonk and Arithmetic Circuit is changed

hovanja2011:


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -133,9 +135,16 @@ instance {-# OVERLAPPING #-} SymbolicData a x => Conditional (Bool (ArithmeticCi
 
 -- TODO: make a proper implementation of Arbitrary
 instance Arithmetic a => Arbitrary (ArithmeticCircuit a) where
-    arbitrary = do
-        let ac = mempty { acOutput = 1} * mempty { acOutput = 2}
-        return ac
+    arbitrary = frequency [
+        (1, return mempty { acOutput = 1}),
```
vlasin, line 139: This only creates a circuit with one input. Let's generalize it to at least up to `100` inputs.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -133,9 +135,16 @@ instance {-# OVERLAPPING #-} SymbolicData a x => Conditional (Bool (ArithmeticCi
...
+        (2, liftM2 (+) arbitrary arbitrary),
+        (2, liftM2 (-) arbitrary arbitrary),
+        (2, liftM2 (*) arbitrary arbitrary),
+        (2, liftM2 (//) arbitrary arbitrary),
+        (2, fromConstant <$> c),
```
vlasin, line 144: `c` could be arbitrary field element.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -133,9 +135,16 @@ instance {-# OVERLAPPING #-} SymbolicData a x => Conditional (Bool (ArithmeticCi
...
+        (2, boolCheckC <$> arbitrary) ]
```
vlasin, line 145: This adds a bool check on a variable. The problem is that if this variable is intermediate, i.e. it is computed using input variables and constants, the circuit is very likely cannot be satisfied. I suggest applying it only to input variables (should be similar to the first case).

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -45,9 +46,11 @@ data Plonk (d :: Natural) (n :: Natural) t = Plonk F F F (Vector n Natural) (Ari
 -- TODO (Issue #25): make a proper implementation of Arbitrary
 instance Arbitrary (Plonk d n t) where
     arbitrary = do
-        let (omega, k1, k2) = getParams 5
+        nP <- integerToNatural <$> chooseInteger (1, 10)
+        let (omega, k1, k2) = getParams nP
         ac <- arbitrary
-        Plonk omega k1 k2 (Vector [1, 2]) ac <$> arbitrary
+        nV <- integerToNatural <$> chooseInteger (1, 100)
+        Plonk omega k1 k2 (Vector [1..nV]) ac <$> arbitrary
```
vlasin, line 53: The vector here gives the indices of public input variables. For now, it should coincide with the circuit inputs.

Note: We are adding private inputs soon, so in the future, this vector will only be a subset of the circuit inputs.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -113,7 +116,8 @@ newtype PlonkWitnessMap d = PlonkWitnessMap (Map.Map Natural F -> (PolyVec F d,
 newtype PlonkWitnessInput = PlonkWitnessInput (Map.Map Natural F)
 -- TODO (Issue #25): make a proper implementation of Show
 instance Show PlonkWitnessInput where
-    show _ = "PlonkWitnessInput"
+    show (PlonkWitnessInput m) = "Witness Input: " ++ show m
+
 -- TODO (Issue #25): make a proper implementation of Arbitrary
 instance Arbitrary PlonkWitnessInput where
```
vlasin, line 122: The problem with this instance is that its definition should be related to the definition of `Arbitrary` for arithmetic circuits. If we allow input variables with indices 1-100 in arbitrary circuits, then this `Map` must contain values for keys 1-100. Keep in mind though that if the circuit's input is constrained to be boolean, then we also need the value to be boolean.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -164,17 +167,34 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
-            c2 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 2}
+    arbitrary = do
+            k <- chooseInteger (0,10)
+            arbitrary' k
+
+arbitrary' :: (Arithmetic a, KnownNat n) => Integer -> Gen (ArithmeticCircuit n a)
+arbitrary' 1 = do
+    oneof [
+        return $ ArithmeticCircuit { acCircuit = mempty {acInput = [integerToNatural 1]}, acOutput = pure 1 }
+        , fromConstant <$> chooseInteger (0, 100)
```
vlasin, line 178: Use `arbitrary` for constants instead of `chooseInteger`.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -114,13 +119,35 @@ newtype PlonkWitnessMap d = PlonkWitnessMap (Map.Map Natural F -> (PolyVec F d,
...
+        arithmeticCircuit :: ArithmeticCircuit n a
+        , witnessInput    :: PlonkWitnessInput
+    }
+
+instance (Arithmetic a, KnownNat n) => Arbitrary (ACandWitness n a) where
+    arbitrary :: Gen (ACandWitness n a)
+    arbitrary = do
+        ac <- arbitrary
+        let keys = inputVariables ac
+        let len = last keys
```
vlasin, line 144: Why not use `ZkFold.Prelude.length`? It seems safer that way.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -44,11 +47,12 @@ type G2 = Point BLS12_381_G2
...
 -- TODO (Issue #25): make a proper implementation of Arbitrary
-instance Arbitrary (Plonk d n t) where
+instance (KnownNat d) => Arbitrary (Plonk d n t) where
     arbitrary = do
-        let (omega, k1, k2) = getParams 5
+        let nP = ceiling @Double . logBase 2.0 . fromIntegral $ value @d
+        let (omega, k1, k2) = getParams nP
         ac <- arbitrary
-        Plonk omega k1 k2 (Vector [1, 2]) ac <$> arbitrary
+        Plonk omega k1 k2 (Vector [1..nP]) ac <$> arbitrary
```
vlasin, line 55: `nP` is not related to the number of inputs. The 4th field of `Plonk` type is the vector of indices of public inputs. It should be a subset of circuit inputs.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -114,13 +118,35 @@ newtype PlonkWitnessMap d = PlonkWitnessMap (Map.Map Natural F -> (PolyVec F d,
 
 newtype PlonkWitnessInput = PlonkWitnessInput (Map.Map Natural F)
 -- TODO (Issue #25): make a proper implementation of Show
-instance Show PlonkWitnessInput where
-    show _ = "PlonkWitnessInput"
--- TODO (Issue #25): make a proper implementation of Arbitrary
 instance Arbitrary PlonkWitnessInput where
```
vlasin, line 121: We cannot properly define this instance without knowing a circuit. Since we also have `ACandWitness`, I suggest removing this one.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -114,13 +118,35 @@ newtype PlonkWitnessMap d = PlonkWitnessMap (Map.Map Natural F -> (PolyVec F d,
...
+        keysLen <- ceiling @Double . logBase 2.0 . fromIntegral <$> chooseInteger (0, 100)
+        let keys = [0..keysLen]
+        values <- vector (integerToInt $ naturalToInteger keysLen) :: Gen [F]
+        let wi = Map.fromList $ zip keys values
+        return (PlonkWitnessInput wi)
+
+instance Show PlonkWitnessInput where
+    show (PlonkWitnessInput m) = "Witness Input: " ++ show m
+
+data (Arithmetic a, KnownNat n) => ACandWitness n a = ACandWitness
```
vlasin, line 132: I suggest renaming it to `ArithmeticCircuitTest`, using `Map` instead of `PlonkWitnessInput`, and moving this definition to the `ArithmeticCircuit` module.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -163,18 +166,59 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
+        newVars = [
+            newInp
+            , newOut
+            ]
+
+        la = frequency $ (1, arbVar) : (1, constant) : Haskell.fmap ((3, ) . return) newVars
+
+        in do
+            l <- la
+            let inp' =  Haskell.max (Haskell.toInteger . Haskell.length $ inputVariables l) inp
```
vlasin, line 198: Can we use `ZkFold.Prelude.length` here?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -163,18 +166,59 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
-instance (Arithmetic a, KnownNat n) => Arbitrary (ArithmeticCircuit n a) where
-    arbitrary = return $ c1 * c2
-        where
-            c1, c2 :: ArithmeticCircuit n a
-            c1 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 1}
-            c2 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 2}
+instance (Arithmetic a, KnownNat n, Arbitrary a) => Arbitrary (ArithmeticCircuit n a) where
+    arbitrary = (ArithmeticCircuit { acCircuit = mempty {acInput = [ 1 ]}, acOutput = pure 1} * ) <$> arbitrary' 0 0 (Haskell.toInteger $ value @n)
+
+arbitrary' :: forall a n . (Arithmetic a, KnownNat n, Arbitrary a, FromConstant a a) => Integer -> Integer ->  Integer -> Gen (ArithmeticCircuit n a)
```
vlasin, line 172: I suggest using `Natural` for nonnegative integers.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -163,18 +166,59 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
 
 -- TODO: make a proper implementation of Arbitrary
-instance (Arithmetic a, KnownNat n) => Arbitrary (ArithmeticCircuit n a) where
-    arbitrary = return $ c1 * c2
-        where
-            c1, c2 :: ArithmeticCircuit n a
-            c1 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 1}
-            c2 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 2}
+instance (Arithmetic a, KnownNat n, Arbitrary a) => Arbitrary (ArithmeticCircuit n a) where
+    arbitrary = (ArithmeticCircuit { acCircuit = mempty {acInput = [ 1 ]}, acOutput = pure 1} * ) <$> arbitrary' 0 0 (Haskell.toInteger $ value @n)
```
vlasin, line 170: I suggest adding `embedVarIndex :: Natural -> ArithmeticCircuit 1 a` to the `ZkFold.Symbolic.Compiler.ArithmeticCircuit.Combinators` module. Use it to create input variables and `embed` to create constants. Using vector variants of those could be even easier.

tests/Tests/NonInteractiveProof/Internal.hs
```diff
@@ -3,17 +3,37 @@
...
+type PlonkSizeBS = 32
+type PlonkBS n = Plonk PlonkSizeBS n ByteString
 
 instance (Show a, Show (Input a), Show (Witness a)) =>
     Show (NonInteractiveProofTestData a) where
     show (TestData a w) = "TestData: \n" ++ show a ++ "\n" ++ show w
 
-instance (NonInteractiveProof a, Arbitrary a, Arbitrary (Witness a)) =>
+instance {-# INCOHERENT #-}
+    (NonInteractiveProof a, Arbitrary a, Arbitrary (Witness a)) =>
```
vlasin, line 27: I propose replacing this definition with a more specific one for KZG.

tests/Tests/Arithmetization.hs
```diff
@@ -20,8 +20,8 @@ import           ZkFold.Symbolic.Types                       (Symbolic)
 propCircuitInvariance :: Arithmetic a => (ArithmeticCircuit 1 a, a, a) -> Bool
 propCircuitInvariance (ac, x, y) =
     let ac' = mapVarArithmeticCircuit ac
-        v   = ac `eval` fromList [(1, x), (2, y)]
-        v'  = ac' `eval` fromList [(1, x), (2, y)]
+        v   = ac `eval` fromList [(0, x), (1, y)]
+        v'  = ac' `eval` fromList [(0, x), (1, y)]
```
vlasin, line 24: Generalize this code according to the new `arbitrary` instance for `ArithmeticCircuit`.

vlasin: Use `ArithmeticCircuitTest` as input.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
```diff
@@ -119,3 +121,9 @@ runInvert r = do
     where
       isZero :: forall a . (Ring a, Eq (Bool a) a, Conditional (Bool a) a) => a -> a
       isZero x = bool @(Bool a) zero one (x == zero)
+
+embedVarIndex :: Arithmetic a => Natural -> ArithmeticCircuit 1 a
+embedVarIndex n = ArithmeticCircuit { acCircuit = mempty { acInput = [ n ]}, acOutput = pure 1}
```
vlasin, line 126: Use `pure n`.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
```diff
@@ -119,3 +121,9 @@ runInvert r = do
...
+
+embedVarIndexV :: (Arithmetic a, KnownNat n) => Natural -> ArithmeticCircuit n a
+embedVarIndexV n = ArithmeticCircuit { acCircuit = mempty { acInput = [ n ]}, acOutput = pure 1}
```
vlasin, line 129: Use `pure n`.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -2,18 +2,22 @@
 {-# LANGUAGE TypeApplications     #-}
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-overlapping-patterns #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
```
vlasin, line 6: This will be removed after specializing the `arbitrary` instance for `NonInteractiveProof`.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -44,11 +48,23 @@ type G2 = Point BLS12_381_G2
...
-        Plonk omega k1 k2 (Vector [1, 2]) ac <$> arbitrary
+        let fullInp = length . inputVariables $ ac
+        vecPubInp <- genSubset (return []) (value @n) fullInp
+        let (omega, k1, k2) = getParams $ value @n
+        Plonk omega k1 k2 (Vector vecPubInp) ac <$> arbitrary
+        where
+            genSubset :: Gen [Natural] -> Natural -> Natural -> Gen [Natural]
+            genSubset arr maxPub maxInp = do
+                len <- length <$> arr
+                case maxPub == len of
```
vlasin, line 62: Use `if`?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -4,29 +4,37 @@
 {-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -Wno-orphans     #-}
+{-# OPTIONS_GHC -Wno-overlapping-patterns #-}
```
vlasin, line 7: Can we remove it?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -164,18 +172,54 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
 
 -- TODO: make a proper implementation of Arbitrary
-instance (Arithmetic a, KnownNat n) => Arbitrary (ArithmeticCircuit n a) where
-    arbitrary = return $ c1 * c2
-        where
-            c1, c2 :: ArithmeticCircuit n a
-            c1 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 1}
-            c2 = ArithmeticCircuit { acCircuit = mempty, acOutput = pure 2}
+instance (Arithmetic a, KnownNat n, Arbitrary a) => Arbitrary (ArithmeticCircuit n a) where
+    arbitrary = (ArithmeticCircuit { acCircuit = mempty {acInput = [ 1 ]}, acOutput = pure 1} * ) <$> arbitrary' 0 0 (value @n)
```
vlasin, line 176: Use `embedVarIndex`?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -164,18 +172,54 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
+
+arbitrary' :: forall a n . (Arithmetic a, KnownNat n, Arbitrary a, FromConstant a a) => Natural -> Natural ->  Natural -> Gen (ArithmeticCircuit n a)
+arbitrary' inp out outMax
```
vlasin, line 179: Use `ArithmeticCircuit`  and recursion depth as input arguments.

tests/Tests/NonInteractiveProof/Internal.hs
```diff
@@ -3,17 +3,36 @@
...
 
-instance (NonInteractiveProof a, Arbitrary a, Arbitrary (Witness a)) =>
-    Arbitrary (NonInteractiveProofTestData a) where
+instance (NonInteractiveProof (KZG c1 c2 t f d), Arbitrary (KZG c1 c2 t f d), Arbitrary (Witness (KZG c1 c2 t f d))) =>
+    Arbitrary (NonInteractiveProofTestData (KZG c1 c2 t f d)) where
     arbitrary = TestData <$> arbitrary <*> arbitrary
+
+instance Arbitrary (NonInteractiveProofTestData (PlonkBS 2) ) where
+    arbitrary = do
+        rbPlonk@(Plonk _ _ _ _ ac _) <- arbitrary :: Gen (PlonkBS 2)
```
vlasin, line 33: To reduce code duplication, I suggest using arbitrary `ArithmeticCircuitTest` here. Then, define `Plonk` and `PlonkWitnessInput` using it. You will need to generate an arbitrary subset in both `NonInteractiveProofTestData` and `Plonk`, so consider making it into a separate function.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -164,18 +171,33 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
+    li <- oneof $ fmap return vars
+    ri <- oneof $ fmap return vars
+    let (l, r) =( ac { acOutput = pure li }, ac { acOutput = pure ri })
+    ac' <- oneof $ fmap return [
+        l + r
+        , l * r
+        , l - r
+        , r - l
+        , l // r
+        , r // l
```
vlasin, line 193: Duplicating division and subtraction makes them twice as likely to be chosen. Since `l` and `r` are already random, maybe we want equal chances for each operation?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
```diff
@@ -119,3 +127,12 @@ runInvert r = do
...
       isZero :: forall a . (Ring a, Eq (Bool a) a, Conditional (Bool a) a) => a -> a
       isZero x = bool @(Bool a) zero one (x == zero)
+
+embedVarIndex :: Arithmetic a => Natural -> ArithmeticCircuit 1 a
+embedVarIndex n = ArithmeticCircuit { acCircuit = mempty { acInput = [ n ]}, acOutput = pure n}
+
+embedVarIndexV :: (Arithmetic a, KnownNat n) => Natural -> ArithmeticCircuit n a
+embedVarIndexV n = ArithmeticCircuit { acCircuit = mempty { acInput = [ n ]}, acOutput = pure n}
+
+getAllVars :: MultiplicativeMonoid a => Circuit a -> [Natural]
```
TurtlePU, line 137: Why is `MultiplicativeMonoid a` constraint needed?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -202,3 +220,26 @@ instance (FromJSON a, KnownNat n) => FromJSON (ArithmeticCircuit n a) where
             let acRNG = mkStdGen 0
                 acCircuit = Circuit{..}
             pure ArithmeticCircuit{..}
+
+type F = Zp BLS12_381_Scalar
```
TurtlePU, line 224: What is the purpose of this type alias?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
hovanja2011: This is necessary for the function `variables`

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -163,19 +167,31 @@ instance {-# OVERLAPPING #-} (SymbolicData a x, n ~ TypeSize a x, KnownNat n) =>
...
+    arbitrary = do
+        k <- integerToNatural <$> chooseInteger (2, 10)
+        let ac = ArithmeticCircuit { acCircuit = mempty {acInput = [1..k]}, acOutput = pure k }
+        arbitrary' ac 10
+
+arbitrary' :: forall a . (Arithmetic a, Arbitrary a, FromConstant a a) => ArithmeticCircuit 1 a -> Natural -> Gen (ArithmeticCircuit 1 a)
+arbitrary' ac 0 = return ac
+arbitrary' ac iter = do
+    let vars = getAllVars . acCircuit $ ac
+    li <- oneof $ fmap return vars
```
TurtlePU, line 180: [`elements`](https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/Test-QuickCheck.html#v:elements) can be used instead of `oneof` in situations like this

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -203,3 +219,25 @@ instance (FromJSON a, KnownNat n) => FromJSON (ArithmeticCircuit n a) where
                 acOutput  = Vector outs
                 acCircuit = Circuit{..}
             pure ArithmeticCircuit{..}
+
+data ArithmeticCircuitTest n a = ArithmeticCircuitTest
```
TurtlePU, line 223: Does this type have any usecases except for testing? Maybe we can move it closer to its actual place of usage in tests along with `mapVar...` function?
@vlasin, does `mapVarArithmeticCircuit` have any intended usages outside of testing?

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -41,14 +43,28 @@ type G2 = Point BLS12_381_G2
...
     arbitrary = do
-        let (omega, k1, k2) = getParams 5
         ac <- arbitrary
-        Plonk omega k1 k2 (Vector [1, 2]) ac <$> arbitrary
+        let fullInp = length . inputVariables $ ac
+        vecPubInp <- genSubset (return []) (value @n) fullInp
+        let (omega, k1, k2) = getParams $ ceiling @Double @Natural $ logBase 2 $ fromIntegral (value @d)
+        Plonk omega k1 k2 (Vector vecPubInp) ac <$> arbitrary
+
+genSubset :: Gen [Natural] -> Natural -> Natural -> Gen [Natural]
```
TurtlePU, line 59: type of the first argument of this function can be safely changed to `[Natural]`

TurtlePU: Or even better to `Set Natural`, as can be inferred from its usage

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
vlasin: Yes, agree.

> @vlasin, does `mapVarArithmeticCircuit` have any intended usages outside of testing?

Nope. Previously, it did, but it wasn't strictly necessary.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -41,14 +43,28 @@ type G2 = Point BLS12_381_G2
...
+genSubset arr maxLength maxValue = do
+    len <- length <$> arr
+    if len == maxLength
+        then arr
+        else do
+            newNat <- integerToNatural <$> chooseInteger (1, toInteger maxValue)
+            let arr' = toList . S.fromList . (newNat : ) <$> arr
```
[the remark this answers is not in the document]
vlasin: You mean, check if we have duplicates, and continue until we don't?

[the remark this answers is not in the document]
TurtlePU: Also there is an algorithm which would produce `maxLength` values which are guaranteed to be distinct:
1. produce `maxLength` values in ranges `[0..maxValue]`, `[0..maxValue -! 1]`, `[0..maxValue -! 2]`, ..., `[0..maxValue -! (maxLength -! 1)]` correspondingly
2. the `i`-th result is the `i`-th random number obtained at the previous step offset by the amount of previous `i` results less than the result
Example: we need 3 numbers in range [0..5], from tossing we get [0,0,0]; the result is [0,1,2] (second `0` is offset by `1`, third `0` is first offset by `1`, then by `1` once again)

But I guess the current algorithm is fast enough so we might not need this, just a thing to keep in mind

TurtlePU (approved):

vlasin (approved):

PR #156, opened 2024-06-27
Multivariate polynomial refactor + Plonk public/private inputs fix

vlasin:
### Multivariate polynomials refactor

I propose to stick, at least for now, with the representation of multivariate polynomials as a "list of coefficients and monomials represented by `Map` from variable indices to powers". With this choice, I improved readability in the related modules and merged some of the Groebner basis-related code there. I will continue this work in another PR.

### Plonk public/private inputs
In the `Plonk` type, we can now specify just the indices of public inputs. See the example in `Test4.hs`.


src/ZkFold/Base/Algebra/Polynomials/Multivariate/Monomial.hs
```diff
@@ -31,47 +18,55 @@ import           Prelude                         hiding (Num (..), drop, filter,
...
-type MonomialAny = M Integer Integer MonomialRepAny
+evalMonomial :: forall i j b .
+    MultiplicativeMonoid b =>
+    Exponent b j =>
+    (i -> b) -> Mono i j -> b
+evalMonomial f (M m) =
+    foldrWithKey (\i j x -> (f i ^ j) * x) (one @b) m
 
-type MonomialBoundedDegree i d = M i Bool (MonomialRepBoundedDegree i d)
+mapVar :: Variable i => [i] -> [i] -> i -> i
```
vks4git, line 47: Could you please add a description of the meaning behind this function? 
Also, is there a way to make it return Maybe? I just don't like the runtime error

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -43,57 +30,61 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate.Monomial
...
     compare (P l) (P r) = compare
         (snd <$> l)
         (snd <$> r)
 
-instance (Arbitrary c, Arbitrary m) => Arbitrary (P c i j m [(c, M i j m)]) where
+instance (Arbitrary c, Arbitrary (Mono i j)) => Arbitrary (Poly c i j) where
     arbitrary = P <$> arbitrary
 
 {-
     In general, `P c i j m p` may define a set of polynomials that is not necessarily a ring.
```
vks4git, line 83: Outdated comment

tests/Tests/Arithmetization/Test4.hs
```diff
@@ -41,13 +43,16 @@ testZKP x ps targetValue =
...
+        setupV  = setupVerify @(PlonkBS N) plonk
         witness = (PlonkWitnessInput inputs, ps)
-        (input, proof) = prove @PlonkBS setupP witness
+        (_, proof) = prove @(PlonkBS N) setupP witness
+
+        -- `one` corresponds to `True`
+        output = plonkVerifierInput $ V.singleton one
 
-    in verify @PlonkBS setupV input proof
+    in verify @(PlonkBS N) setupV output proof
```
zlonast, line 55: Maybe we should name it `none`?

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Monomial.hs
vlasin: I changed the function signature and added a comment. For now, I left the runtime error.

tests/Tests/Arithmetization/Test4.hs
vlasin: Do you mean `output`? We want to verify that the circuit returns symbolic `True` represented by `Zp 1`.

vlasin: There is a difference between inputs to the circuit and inputs to the protocol. The latter actually includes circuit outputs.

zlonast: I think something is wrong with `output` in input, maybe `noneInput` or `onlyOutput` ?

zlonast (dismissed):

tests/Tests/Arithmetization/Test4.hs
vlasin: Change to `circuitOuputsTrue`.

PR #158, opened 2024-06-29
Minor Plonk refactor

vlasin:
- [x] Add `PlonkConstraint` type for a single vanilla Plonk constraint.
- [x] Add `PlonkRelation` type for the relation that is being verified.
- [x] Removed "variable mapping" from the protocol (it was not necessary, actually).
- [x] Add a class to test the compatibility of two non-interactive proof protocols. We take `prove` from one protocol and `verify` from the other.


src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
```diff
@@ -0,0 +1,74 @@
...
+import           ZkFold.Base.Algebra.Basic.Number
+import           ZkFold.Base.Algebra.Basic.Permutations       (Permutation, fromCycles, mkIndexPartition)
+import           ZkFold.Base.Algebra.Polynomials.Multivariate (evalMonomial, evalPolynomial, var)
+import           ZkFold.Base.Algebra.Polynomials.Univariate   (PolyVec, toPolyVec)
+import           ZkFold.Base.Data.Vector                      (Vector, fromVector)
+import           ZkFold.Base.Protocol.ARK.Plonk.Constraint    (PlonkConstraint (..), toPlonkConstraint)
+import           ZkFold.Prelude                               (replicate)
+import           ZkFold.Symbolic.Compiler
+
+type PlonkPolyDegree n = 2 ^ (Log2 n + 1)
```
zlonast, line 21: I think this is cool, but I don't understand where it could be used in type level.

zlonast (approved):

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
vlasin: Yep, forgot to delete it. Since we do not know the circuit size at compile time, we cannot determine a proper Plonk polynomial degree at compile time.

PR #161, opened 2024-07-04
Use associated types for `BinaryExpansion` output

TurtlePU:
As there is always a single (the most efficient and the most correct) implementation of `BinaryExpansion` for an input type, I propose to turn the second parameter into an associated type. Most importantly, this way usage of `Arithmetic a` constraint does not force to turn on an `UndecidableInstances` language extension.


vlasin:
Sounds good to me.

vlasin (approved):

PR #166, opened 2024-07-05
Add `FieldElementData` class

vlasin:
### Motivation

In the context of computations subject to verification using a proving system that is based on prime fields, we need to "serialize" data into structures containing field elements. Moreover, we need "numeric" serialization, which produces a structure containing concrete values, as well as "symbolic" serialization, which produces a structure containing variables (aka symbols).

In this PR, we

- [x] introduced `FieldElementData`
- [x] extended hashing with `MiMC` algorithm to all types of class `FieldElementData`

NOTE: we derived instances of `FieldElementData` using `SymbolicData` typeclass. When the latter is removed, we expect the code to migrate into the corresponding  `FieldElementData` instances.


echatav (approved):
LGTM

PR #176, opened 2024-07-12
Small refactoring plonk

zlonast:
- [x] move `bls12_381_G1_generator` and `bls12_381_G2_generator` to on-chain code
- [x] add testAttackSafeOneInputZKP without output in public constant (the test should fail under attack when we pass false)
- [x] add testSafeOneInputZKP without output in public constant, but with additional constraint `y - 1 = 0` (it should be safe)
- [x] add safeZero for `y - 1 = 0`
- [x] add Generic for PlonkPermutation


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Combinators.hs
```diff
@@ -120,6 +121,11 @@ desugarRange i b
           | c == zero = ($ j) * (one - ($ k))
           | otherwise = one + ($ k) * (($ j) - one)
 
+safeZero :: (Arithmetic a, Traversable f) => ArithmeticCircuit a f -> ArithmeticCircuit a f
+safeZero r = circuitF $ do
+    is' <- runCircuit r
+    for is' $ \i -> newConstrained (\x j -> (x i - one) * x j) (\x -> x i - one)
```
TurtlePU, line 127: As written, $x_j = x_i - 1$, so the constraint simplifies to $(x_i - 1) x_j = x_j^2$ which can be simplified to just $x_j$ or `constraint (\x -> x i - one)` altogether without introducing new variable.
Is there a typo or am I misunderstanding something?

TurtlePU (approved):
Need to fix conflicts in Arithmetization test No.4

TurtlePU (approved):

vlasin (approved):

Issue #154, opened 2024-06-27
Numerical Hierarchy Proposal


echatav:
Trying to resolve our concepts of `Arithmetic` fields, `Symbolic` fields, our existing numerical hierarchy, Haskell's numerical hierachy, and the mathematics of semirings leads me to this numerical hierarchy proposal sketch:

```Haskell
-- Examples of a symbolic fields should include:
-- Prime p => Zp p
-- SymbolicField a => i -> a
-- PrimeField a => ArithmeticCircuit i Par1 a
type SymbolicField a = (Field a, FiniteOrder a, Trichotomy a, Degree a)

-- Prime fields should be `Prime p => Zp p` or `Zp` newtypes.
type PrimeField a = (SymbolicField a, SemiIntegral a)

class AdditiveSemigroup a where
  (+) :: a -> a -> a

class AdditiveSemigroup a => AdditiveMonoid a where
  zero :: a
  combineNatural :: [(Natural,a)] -> a
  scaleNatural :: Natural -> a -> a

class AdditiveMonoid a => AdditiveGroup a where
  negate :: a -> a
  (-) :: a -> a -> a
  combineInteger :: [(Integer,a)] -> a
  scaleInteger :: Integer -> a -> a

class MultiplicativeSemigroup a where
  (*) :: a -> a -> a

class MultiplicativeSemigroup a => MultiplicativeMonoid a where
  one :: a -> a -> a
  powNatural :: a -> Natural -> a

class MultiplicativeMonoid a => MultiplicativeGroup a where
  (/) :: a -> a -> a
  powInteger :: a -> Integer -> a

class From x a where from :: x -> a

type Semiring a = (AdditiveMonoid a, MultiplicativeMonoid a, From Natural)
type Ring a = (AdditiveGroup a, MultiplicativeMonoid a, From Integer)
type Field a = (AdditiveGroup a, MultiplicativeGroup a, From Rational)

class Semiring a => FiniteOrder a where
  order :: a -> Natural

class Semiring a => Degree a where
  degreeBits :: a -> [a]
  fromBits :: [a] -> a

class Degree a => SemiEuclidean a where
  divMod :: a -> a -> (a,a)
  euclidBit :: a -> a -> (a,a,a,a)

class (Ring a, SemiEuclidean a) => Euclidean a where
  quotRem :: a -> a -> (a,a)
  euclid :: a -> a -> (a,a,a)

class Semiring a => Discrete a where
  equal :: a -> a -> a

[...]

PR #177, opened 2024-07-12
Proof-of-Concept: Structured I/O, Numeric Hierarchy with (Semi)Euclidean + (Semi)Integral, Optimized Constants, Map-of-Maps multivariate polynomials

echatav:
Proof-of-Concept draft reference demonstrating:
* Structured Circuit I/O
* Numeric/Algebraic typeclass hierarchy with
  * (`Semi`)`Euclidean` & (`Semi`)`Integral`
  * optimizable scalar `combine` and `evalMono`mial methods
  * `Mod int n` parametrizable by a `SemiIntegral int` and a `KnownNat n` or `Prime n`.
* Map-of-maps definition of multivariate `Poly`nomials
* Optimization of constant variables.


vlasin:
We've recently added a new type of constraints called _range constraints_ to the `Circuit` type. An input or output variable can be constrained to an interval of the form `[0, a]`. I think it would make sense to consider it in this draft.

Additionally, we will have constraints of the form `(x_1, ..., x_k) \in T` sometime soon. Just a heads up.

src/ZkFold/Symbolic/Base/Num.hs
```diff
@@ -0,0 +1,743 @@
...
+-- (SemiIntegral int, Prime p) => PrimeField (int `Mod` p)
+-- and newtypes of int `Mod` p.
+-- p = 2 is ruled out to allow/require trichotomy.
+type PrimeField x =
+  ( Modular x
+  , Finite x
+  , Symbolic x x
+  )
+
+class AdditiveMonoid a where
```
TurtlePU, line 103: I agree that we might not need the `AdditiveSemigroup` (and `MultiplicativeSemigroup`) base classes as we currently have no usecase for them (and all the datatypes are at least monoids)

src/ZkFold/Symbolic/Base/Num.hs
```diff
@@ -0,0 +1,743 @@
...
+  , From Natural a
+  , From Integer a
+  , From a a
+  , Scalar Natural a
+  , Scalar Integer a
+  , Scalar a a
+  , Exponent Natural a
+  )
+
+type Field a =
```
TurtlePU, line 166: I still don't agree that `Field` is just an `AdditiveGroup` + `MultiplicativeGroup`, the group of invertible elements of a field is a separate object with its own size (in case we're talking about finite objects), so to treat the whole field as a multiplicative group is not correct IMO

src/ZkFold/Symbolic/Base/Num.hs
```diff
@@ -0,0 +1,743 @@
...
+instance (SemiIntegral int, Prime p, KnownNat (Log2 (p - 1) + 1), 3 <= p)
+  => Symbolic (Mod int p) (Mod int p)
+
+-- Scalar ------------------------------------------------------------------
+class (Semiring c, AdditiveMonoid a)
+  => Scalar c a where
+    scale :: c -> a -> a
+    default scale :: c ~ a => c -> a -> a
+    scale = (*)
+    combine :: [(c,a)] -> a
```
TurtlePU, line 677: A useful operation to be added to perform some operations faster via Fiat-Shamir trick, as explained to me by @echatav

TurtlePU:
I think we salvaged the ideas from this proposal in #154, #214 and in #217, so we can close this. @echatav, what do you think?

echatav:
I just want to keep it open as a reference, not to merge. There are still lots of things here as proof of concepts that are relevant to proposals that I want to merge back in their own PRs. I will add a label if it helps.

PR #187, opened 2024-07-23
Try to unify benchmarking, prototype witness generation measurement

iokasimov:
According to measurements, the bigger circuit become the less time witness generation takes.

I've tried to generalize code to fit various circuits.


bench/BenchAC.hs
```diff
@@ -32,115 +40,171 @@ evalBS (ByteString xs) = eval xs M.empty
...
+
+reverseListCircuit :: forall t n . Vector n t -> Vector n t
+reverseListCircuit = unsafeToVector . reverse . fromVector
+
+benchCircuit :: forall n p .
+    KnownNat n =>
+    String ->
+    IO (ByteString n (ArithmeticCircuit (Zp p))) ->
+    Benchmark
+benchCircuit desc crct = env (crct >>= evaluate . force) $ bench title . nf evalBS where
```
TurtlePU, line 75: What does this function benchmark?

bench/BenchAC.hs
```diff
@@ -32,115 +40,171 @@ evalBS (ByteString xs) = eval xs M.empty
...
+--       , benchCircuit "Adding ByteStrings" $ additionCircuit @128 @BLS12_381_Scalar
+--       , benchCircuit "Adding ByteStrings" $ additionCircuit @256 @BLS12_381_Scalar
+--       , benchCircuit "Adding ByteStrings" $ additionCircuit @512 @BLS12_381_Scalar
+--       ]
+
+-- mainSumWitnnessGeneration :: IO ()
+-- mainSumWitnnessGeneration = do
+
+--   defaultMain
+--       [ benchWitnessGeneration "Adding ByteStrings" $ toFieldElements @(ArithmeticCircuit F) <$> additionCircuit @32 @BLS12_381_Scalar
```
TurtlePU, line 205: Maybe we can include witness generation into the benchmark as well, just to make it be seen in case something goes wrong?

vlasin: The general idea is to have separate benchmarks for compilation and evaluation that would share a collection of functions on which we run them.

TurtlePU: So I guess the benchmarks have to be split in different modules? How should this work with the cabal's benchmarking framework

bench/BenchAC.hs
iokasimov: @TurtlePU Whole circuit evaluation. I just generalized some functions for benchmarking that looked almost the same.

vlasin:
A path to merge:

- [ ] `benchAC`  should be a benchmark of circuit size;
- [ ] The list of circuits to benchmark should be moved to a separate module as it is shared between different benchmarks;
- [ ] Code repetitions should be refactored.

Leaving finishing this PR to @TurtlePU and his team.

TurtlePU:
@vlasin, ready for review!

Changes made:

* there are only 2 benchmarks now: `BenchPolyMul` for benchmarking polynomial multiplication, and `BenchCompiler` for benchmarking compilation of Symbolic smart contracts.
* `zkfold-base-examples` test is repurposed into the _library_ of examples, which are collected in `examples/Examples` folder for ease of viewing. All of them are collected into unevaluated circuit thunks stored in an `examples` list defined in `ZkFold.Symbolic.Examples` module to be reused as sample benchmarks and sample tests.
* Compiler benchmark then benchmarks compilation and witness calculation of these thunks, and finally stores the circuit metrics (# of constraints, # of variables, # of range checks) in `stats/` folder. They are checked out to be later compared against; `tasty-golden` performs these checks automatically.
* If any of our targets is run with `--enable-profiling` Cabal flag, they in addition generate the `.prof` file which can be viewed in https://speedscope.app. Compiler benchmark flags are specified in a way that _all_ functions are reported in the flamegraph, this can be tweaked with `-fprof-*` GHC options.
* Github CI workflows are tweaked so that all targets are build, not only `-base` and `-test` and (here gone) `-examples`

vlasin (approved):

PR #189, opened 2024-07-23
context is now `(Type -> Type) -> Type`

TurtlePU:
This PR performs the migration from Natural-indexed contexts to functor-indexed ones.
Note that the only change is the type of contexts: `SymbolicData` and friends stay the same; `Symbolic` class is unchanged; symbolic datatypes just use `Vector`s internally.
Other changes will be performed in the future PRs.


vlasin (approved):

echatav (approved):
:shipit:

PR #195, opened 2024-07-28
Complete Symbolic definition

TurtlePU:
1. Added `MonadCircuit` as a superclass to the (soon to be deprecated) `MonadBlueprint`. The latter now only contains the `runCircuit`, every other member has been moved to `MonadCircuit`.
2. Completed `Symbolic` class definition with a `symbolicF` function.
3. Various utilities and QoL improvements to simplify working with new `Symbolic` class.
4. Docs

### DIFF
--- a/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
+++ b/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DeriveAnyClass   #-}
 {-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeOperators    #-}
 
 module ZkFold.Symbolic.Compiler.ArithmeticCircuit.Internal (
         ArithmeticCircuit(..),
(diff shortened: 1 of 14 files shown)


src/ZkFold/Symbolic/MonadCircuit.hs
```diff
@@ -0,0 +1,92 @@
...
+-- * Variables with either different constraints or different witnesses are different;
+-- * There is an order in which witnesses can be generated;
+-- * Constraints never reference undefined variables.
+--
+-- However, DSL does NOT provide the following guarantees (yet):
+--
+-- * That provided witnesses satisfy the provided constraints. To check this,
+--   you can use 'ZkFold.Symbolic.Compiler.ArithmeticCircuit.checkCircuit'.
+-- * That introduced constraints are supported by the zk-SNARK utilized for later proving.
+class Monad m => MonadCircuit i a m | m -> i, m -> a where
```
vlasin, line 67: I suggest adjusting the description of `newRanged` and `newConstrained` to emphasize that they also add new constraints besides just declaring a new variable.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -106,8 +107,45 @@ instance (Eq a, MultiplicativeMonoid a) => Package (ArithmeticCircuit a) where
...
+
+-------------------------------- MonadCircuit instance ------------------------------
+
+instance Arithmetic a => MonadCircuit Natural a (State (Circuit a)) where
+    newRanged upperBound witness = do
+        let s   = sources @a witness
+            -- | A wild (and obviously incorrect) approximation of
+            -- x (x - 1) ... (x - upperBound)
+            -- It's ok because we only use it for variable generation anyway.
+            p i = var i * (var i - fromConstant upperBound)
```
vlasin, line 124: We should think of a safer way to do it. `p` could be used to define a constraint `x \in {0, upperBound}`.

TurtlePU: Oh, definitely, didn't think about set inclusion constraints when writing this

vlasin (approved):

PR #202, opened 2024-08-02
check that tests pass on main

echatav:


vlasin:
Any idea how we could fix this CI fail?

TurtlePU:
My guess is that Github cancels the job after it considers the job "idle" because it writes nothing to the console long enough
So the solution is to make the test run faster

TurtlePU:
This, or either the whole time dedicated to the job is out, so to make it pass we have to compile faster: optimize UInt compilation times, split up the package etc.

Or increase the time limit of the GitHub Action

vlasin:
I did not find any other limitations for a public repository. The runner machine has 16 GB RAM and 14 GB SSD.

echatav:
Also, the particular tests that fail/timeout on CI seem to run not _that_ slow on my machine. So I'm at a loss.

TurtlePU:
Let's reorder the tests and see what happens :D

TurtlePU:
> However, the failed run above took only 38 minutes and was terminated during the `pairing` test.

Actually, the run took almost 43 minutes, 5 of which were spent on installing GHC and building the project

vlasin:
1. Is it necessary to install GHC every time?
2. For properties that take long time to check, we could reduce the number of test runs.
3. If we really need to, we could split tests into several test suites and chain workflows to reset the timer.
4. Finally, in the future we might want to split the package. However, as already discussed, right now is maybe too soon for that.

TurtlePU:
> 1. Is it necessary to install GHC every time?

I hope we can cache the installation to preserve ot between runs. @echatav is that so?

>     2. For properties that take long time to check, we could reduce the number of test runs.

True. The most time-consuming tests are:
* pairing spec -- will become faster with Rust FFI
* Symbolic tests -- I have an idea how to avoid constructing circuits explicitly but still check that function definitions are correct, this should speed things up. Also lookup arguments should help.

>     3. If we really need to, we could split tests into several test suites and chain workflows to reset the timer.

Sure, if we really have to... 

>     4. Finally, in the future we might want to split the package. However, as already discussed, right now is maybe too soon for that.

Totes.

vlasin:
> * pairing spec -- will become faster with Rust FFI

Rust FFI is now in another repo. This one will only contain a naive Haskell implementation.

TurtlePU:
New job failed in 45 mins
I guess the case is clear, 45 min time limit is not enough
My machine is also 16Gb of RAM, but has an i5 processor and has more SSD. Takes a whole lot longer than 45 mins to run the whole test suite, so I think it is fair that it times out here too

So we indeed have to think about how to run tests faster (and what to test in given time).

vlasin:
Ok, given that we need to bring the time down significantly, maybe it makes sense to focus on benchmarks and profiling first.

- For things like elliptic curve operations, pairing and Plonk protocol, we will rely on FFI implementations for efficiency, so we can just reduce the number of runs.
- For Symbolic compilation and evaluation, we want to optimize as much as possible.

PR #205, opened 2024-08-05
Merged Conditional into a single instance

TurtlePU:
Merges two conditional instances into one, getting rid of most `Conditional` constraints in signatures (except for one in `Ledger.Types`, as here an instance of `SymbolicData` for `List` is required)

Also a question about circuit for conditionals: does this really introduce a Plonk constraint? The resulting constraint actually mentions 4 variables.


vlasin:
> Also a question about circuit for conditionals: does this really introduce a Plonk constraint? The resulting constraint actually mentions 4 variables.

Hmm, I think we missed the moment when the constraint became non-Plonk. Vanilla Plonk constraints only have 3 variables.

TurtlePU:
@vlasin, changed constraints to be Plonk

vlasin (approved):

echatav (approved):
LGTM

PR #207, opened 2024-08-05
Migrated Ord to Symbolic API

TurtlePU:
Migrated basic `Ord` instances to Symbolic API.
Motivation for adding `Haskell.Ord a` to `Arithmetic a` is that this is actually implied anyway bc we can convert `a` to `Natural` via `BinaryExpansion a` and `Haskell.Eq a`


echatav (approved):
LGTM

PR #211, opened 2024-08-07
Symbolic `FieldElement`

TurtlePU:
Migrated `FieldElement` to `Symbolic` API, removed algebraic instances of contexts

Note: in TrichotomyField instance, the constraint might not be Plonk.


echatav (approved):
:shipit:

PR #213, opened 2024-08-08
+ bn254

TurtlePU:
This PR adds bn254 elliptic curve as a viable backend for zk proving schemes @ ZkFold.


TurtlePU:
@zlonast, @vlasin, maybe we merge this and save pairing refactoring for later?

zlonast:
> @zlonast, @vlasin, maybe we merge this and save pairing refactoring for later?

Do I understand correctly that curve pairing does not work now?
Anyway, i would like this pull request to be accepted as soon as possible. 👀

zlonast (dismissed):

TurtlePU:
> Do I understand correctly that curve pairing does not work now?

For BLS pairing still works, for BN it does not work yet, the algorithm for BN curves is slightly different so we need to refactor a bit

vlasin:
Apparently, the point compression algorithm for BLS, which is currently in `main`, does not coincide with the point compression used in Cardano (tagging @charrsky). With that in mind, I have two questions:

- Is there a standard algorithm for that?
- Could you check if it is the same algorithm that is used here: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0381?

echatav:
Here is the [documentation of the serialization format](https://github.com/supranational/blst?tab=readme-ov-file#serialization-format) used in that library and what I modeled the serialization code in zkfold-base upon:

zlonast:
> Here is the [documentation of the serialization format](https://github.com/supranational/blst?tab=readme-ov-file#serialization-format) used in that library and what I modeled the serialization code in zkfold-base upon. 

Same library as [cardano-crypto](https://github.com/IntersectMBO/cardano-base/blob/9b633d49ce5390591897576c97301319b936f6ed/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs), but for some reason our implementation behaves differently 🤔

echatav:
If I had to guess where the code is going wrong it would be where it's most fragile. It has to do a lot of logic and work converting between its arbitrary size `Integer` representation of field elements and their fixed size representations. For BLS12, Fq is 48 bytes which is best represented in Haskell using the [`LargeKey` type](https://hackage.haskell.org/package/largeword-1.2.5/docs/Data-LargeWord.html):

```Haskell
type Word48 = LargeKey Word16 Word32
```
Maybe it would help to simplify the code using this type?

TurtlePU:
From link provided by @echatav:

> The third-most significant bit is set if (and only if) this point is in compressed form and it is not the point at infinity and its y-coordinate is the lexicographically largest of the two associated with the encoded x-coordinate.

But we check the third bit in uncompressed rep, too. Seems like algorithms are different after all

TurtlePU:
My hypothesis is that the "lexicographical order" from the spec @echatav sent is different from the order on points we use, but while exploring the source code I struggle to understand the assembly they use to compute the "big y" bit: https://github.com/supranational/blst/blob/master/src/asm/mul_mont_384-armv8.pl#L1627

Can you tell if I'm looking in the right place or the error persists even with uncompressed points?

vlasin (changes requested):
Let's not rush this. We need to

- [ ] figure out the correct generic compression algorithm (must agree with `blst`),
- [ ] make pairing work for both BLS and BN curves.

vlasin (approved):

Issue #113, opened 2024-05-21
Incorporate all necessary types for zkFold Symbolic smart contracts into the `Symbolic` constraint


vlasin:
We want a typical zkFold Symbolic smart contract to have the following signature:
`contract :: (Symbolic t, Arithmetic a) => t Tx a -> t Input a -> t Bool a`

Here `t` could be the identity or `ArithmeticCircuit`, and `a` is the chosen finite field.

TurtlePU:
So, after #123, we can view Symbolic datatypes as such:

* While running with `Zp p` as base &mdash; `Vector typeSize (Zp p)`
* While running with arithmetic circuits as base &mdash; `ArithmeticCircuit typeSize (Zp p)`

After replacing natural indexing with containers as is currently done in #121 and generalizing from `Zp p` to generic `Arithmetic a`, we get

* `container a` for pure backend
* `ArithmeticCircuit container a` for AC backend

This motivates introducing the following typeclass:
```haskell
class Arithmetic a => Symbolic a t where
  symbolic ::
    (SymbolicData a f, SymbolicData a g) =>
    (forall i m . MonadBlueprint i a m => f i -> m (g i)) ->
    (f a -> g a) -> t f a -> t g a
```

So the user-end code will look like this:
```haskell
instance Symbolic a t => AdditiveSemigroup (t (UInt n) a)

contract :: Symbolic a t => t Tx a -> t Input a -> t Bool a
```

With additional type synonyms we can get:
```
type (~>) f g = forall a t . Symbolic a t => t f a -> t g a
contract :: Tx ~> Input ~> Bool
```

TurtlePU:
Actually, there is also a useful typeclass hierarchy preceding `Symbolic`:
```haskell
class HFunctor t where
  hmap :: (forall a . f a -> g a) -> t f a -> t g a

class HFunctor t => HApplicative t where
  hpure :: (forall a . f a) -> t f a
  hliftA2 :: (forall a. f a -> g a -> h a) -> t f a -> t g a -> t h a

class Lift t where
  lift :: Functor f => f a -> t f a

class (HApplicative t, Lift t) => Symbolic a t where
  symbolic ::
    (SymbolicData a f, SymbolicData a g) =>
    (forall i m . MonadBlueprint i a m => f i -> m (g i)) ->
    (f a -> g a) -> t f a -> t g a
```

TurtlePU:
@echatav
```haskell
class SFunctor f where
[...]

PR #214, opened 2024-08-09
Structured i/o

echatav:
First pass at merging structured inputs.

Still needs
* optimization of constant variables
* other cleanups


src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
```diff
@@ -38,11 +38,14 @@ toPlonkRelation :: forall n l a .
...
     => Scale a a
     => Vector l Natural
-    -> ArithmeticCircuit a Par1
+    -> ArithmeticCircuit a (Vector l) Par1
     -> Maybe (PlonkRelation n l a)
 toPlonkRelation xPub ac0 =
     let ac = desugarRanges ac0
-        evalX0 = evalPolynomial evalMonomial (\x -> if x == 0 then one else var x)
+
+        varF (NewVar ix) = if ix == 0 then one else var (ix + value @l)
```
TurtlePU, line 46: So after this PR we need to decide which input is the constant 1, and `NewVar 0` looks like a good choice.

src/ZkFold/Symbolic/Cardano/Types/Address.hs
```diff
@@ -23,8 +24,9 @@ deriving instance (Haskell.Eq (ByteString 4 context), Haskell.Eq (ByteString 224
 
 deriving instance HApplicative context => SymbolicData context (Address context)
 
-deriving via (Structural (Address CtxCompilation))
-         instance Eq (Bool CtxCompilation) (Address CtxCompilation)
+deriving via (Structural (Address (CtxCompilation i)))
```
TurtlePU, line 27: Will generalize it to `Symbolic c` after this PR

src/ZkFold/Symbolic/Cardano/Types/Output.hs
```diff
@@ -39,13 +40,15 @@ deriving instance
...
-            ( ts ~ TypeSize CtxCompilation (Output tokens datum CtxCompilation)
+            ( ts ~ TypeSize (CtxCompilation i) (Output tokens datum (CtxCompilation i))
             , 1 <= ts
             , KnownNat tokens
-            , KnownNat (TypeSize CtxCompilation (Value tokens CtxCompilation))
-            ) => Eq (Bool CtxCompilation) (Output tokens datum CtxCompilation)
+            , KnownNat (TypeSize (CtxCompilation i) (Value tokens (CtxCompilation i)))
+            , Ord (Rep i)
+            , Representable i
+            ) => Eq (Bool (CtxCompilation i)) (Output tokens datum (CtxCompilation i))
```
TurtlePU, line 51: Needs to be generalized to `Symbolic c` after this PR

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
```diff
@@ -48,52 +50,50 @@ import           ZkFold.Prelude                                         (length)
...
                                                                          witnessGenerator)
 import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.Map
 
 --------------------------------- High-level functions --------------------------------
 
 -- TODO: make this work for different input types.
-applyArgs :: ArithmeticCircuit a f -> [a] -> ArithmeticCircuit a f
-applyArgs r args =
-    (execState (apply args) r {acOutput = U1}) {acOutput = acOutput r}
+-- applyArgs :: ArithmeticCircuit a (i :*: j) o -> i a -> ArithmeticCircuit a j o
```
TurtlePU, line 60: Turns out that this partial application function was not used anywhere but in tests, but there a complete application (`eval`) was used actually.

Also, for partial application purposes, `Symbolic` functions + `FromConstant` instances work better and more intuitively in my opinion, maybe we use it from here on out?

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
echatav: Yes, after we optimize constants we won't need a special variable for `one` or any other constant.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
```diff
@@ -27,13 +26,13 @@ import           ZkFold.Symbolic.Data.FieldElement                      (FieldEl
 
 ------------------------------------- Instances -------------------------------------
 
-instance (Arithmetic a, Arbitrary a) => Arbitrary (ArithmeticCircuit a Par1) where
+instance (Arithmetic a, Arbitrary a, Arbitrary (Rep i), Haskell.Ord (Rep i), Representable i, Haskell.Foldable i) => Arbitrary (ArithmeticCircuit a i Par1) where
     arbitrary = do
-        k <- integerToNatural <$> chooseInteger (2, 10)
-        let ac = mempty { acInput = [1..k], acOutput = pure k }
+        outVar <- InVar <$> arbitrary
```
TurtlePU, line 31: With `i ~ Vector n`, this would cause problems because the generated `outVar` most probably would be out of bounds

My proposed fix would be to change the `Rep (Vector n)` to `Zp n`

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -45,63 +50,74 @@ import           ZkFold.Base.Algebra.Polynomials.Multivariate (Mono, Poly, evalM
...
+deriving stock instance Eq (Rep i) => Eq (Var i)
+deriving stock instance Ord (Rep i) => Ord (Var i)
+deriving instance NFData (Rep i) => NFData (Var i)
+
+witnessGenerator :: ArithmeticCircuit a i o -> i a -> Map Natural a
 witnessGenerator circuit inputs =
-  let srcs = acWitness circuit
-      witness = ($ witness) <$> (srcs `union` fmap const inputs)
-   in witness
+    fmap ($ inputs) (acWitness circuit)
```
TurtlePU, line 93: Because of this simpler setup, witness computation takes (exponentially?) more time because now each new value in a witness recomputes all other values it depends on, they do the same...
The same loeb trick as was here previously is achievable with `acWitness :: Map Natural (i a -> Map Natural a -> a)`.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -179,20 +196,22 @@ toField :: Arithmetic a => a -> VarField
...
 -- TODO: Remove the hardcoded constant.
-toVar :: Arithmetic a => [Natural] -> Constraint a -> Natural
+toVar :: Arithmetic a => [Var i] -> Constraint a i -> Natural
 toVar srcs c = force $ fromZp ex
     where
         r  = toZp 903489679376934896793395274328947923579382759823 :: VarField
         g  = toZp 89175291725091202781479751781509570912743212325 :: VarField
-        v  = (+ r) . fromConstant
+        varF (NewVar w) = w
+        varF (InVar _)  = 0
```
TurtlePU, line 205: This one might turn out to be problematic if the only variable we depend on is an input variable.

All in all, I think we should change our value generation algorithm to generate it from witness instead: the basic idea is to view witness as a syntax tree which we can Merkle-hash.

P.S. Then we would need `Rep i` to be hashable; this in turn means that we might want to migrate to hashmaps in arithmetic circuit to reduce the number of constraints we impose on `i`, maybe?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -203,43 +222,56 @@ addVariable x = do
...
+exec ac = eval ac U1
 
 -- | Applies the values of the first `n` inputs to the arithmetic circuit.
 -- TODO: make this safe
-apply :: [a] -> State (ArithmeticCircuit a U1) ()
-apply xs = do
-    inputs <- gets acInput
-    zoom #acInput . put $ drop (length xs) inputs
-    zoom #acWitness . modify . union . fromList $ zip inputs (map const xs)
+apply :: (Eq a, Field a, Ord (Rep j), Scale a a, FromConstant a a, Representable i) => i a -> ArithmeticCircuit a (i :*: j) U1 -> ArithmeticCircuit a j U1
```
TurtlePU, line 261: Also `Scale a a` and `FromConstant a a` turned up all over the place. Might as well add them to `Arithmetic a` constraint in the end

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Map.hs
```diff
@@ -6,58 +6,56 @@ module ZkFold.Symbolic.Compiler.ArithmeticCircuit.Map (
...
-instance (Arithmetic a, Arbitrary a, Arbitrary (ArithmeticCircuit a Par1)) => Arbitrary (ArithmeticCircuitTest a Par1) where
-    arbitrary :: Gen (ArithmeticCircuitTest a Par1)
+instance (Arithmetic a, Arbitrary a, Arbitrary (ArithmeticCircuit a i Par1), Traversable i, Representable i) => Arbitrary (ArithmeticCircuitTest a i Par1) where
+    arbitrary :: Gen (ArithmeticCircuitTest a i Par1)
     arbitrary = do
         ac <- arbitrary
-        let keysAC = acInput ac
-        values <- vector . integerToInt . naturalToInteger . length  $ keysAC
-        let wi = fromList $ zip keysAC values
+        wi <- for acInput $ \_ -> arbitrary
```
TurtlePU, line 41: Maybe `Arbitrary (i a)` is better for this in the end

src/ZkFold/Symbolic/Data/FieldElement.hs
```diff
@@ -102,30 +97,3 @@ instance Symbolic c => BinaryExpansion (FieldElement c) where
   fromBinary bits =
     FieldElement $ symbolicF bits (Par1 . foldr (\x y -> x + y + y) zero)
       $ fmap Par1 . horner . fromVector
-
-instance Symbolic c => DiscreteField' (FieldElement c) where
```
TurtlePU, line 106: as `Discrete` and `Trichotomy` stuff is not needed, maybe we can then remove them from our typeclass hierarchy

tests/Tests/ArithmeticCircuit.hs
```diff
@@ -47,20 +46,20 @@ specArithmeticCircuit' = hspec $ do
         it "has one" $ correctHom0 @a one
         it "inverts nonzero correctly" $ correctHom1 @a finv
         it "inverts zero correctly" $ correctHom0 @a (finv zero)
-        it "checks isZero(nonzero)" $ \(x :: a) ->
-          let Bool (r :: ArithmeticCircuit a Par1) = isZero $ FieldElement (embed x)
-           in checkClosedCircuit r .&&. exec1 r === bool zero one (x Haskell.== zero)
-        it "checks isZero(0)" $
-          let Bool (r :: ArithmeticCircuit a Par1) = isZero (zero :: FieldElement (ArithmeticCircuit a))
-           in withMaxSuccess 1 $ checkClosedCircuit r .&&. exec1 r === one
+        -- it "checks isZero(nonzero)" $ \(x :: a) ->
```
TurtlePU, line 49: We will bring back tests for `isZero` sometime later

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
TurtlePU: I remember @vlasin saying that this variable is still needed in zk proving algorithms somehow, or do I remember it wrong?

TurtlePU (changes requested):
Needs to be addressed before merging:
* Fix `Representable (Vector n)` instance to be exception-safe
* Add loeb trick to witness calculation

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
echatav: I remember the opposite 🤷

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Instance.hs
echatav: I made this change.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -136,33 +160,26 @@ sources = runSources . ($ Sources @a . S.singleton)
...
 ----------------------------------- Circuit monoid ----------------------------------
 
-instance (Eq a, o ~ U1) => Semigroup (ArithmeticCircuit a o) where
+instance (Eq a, o ~ U1) => Semigroup (ArithmeticCircuit a i o) where
     c1 <> c2 =
         ArithmeticCircuit
            {
                acSystem   = acSystem c1 `union` acSystem c2
             ,  acRange    = acRange c1 `union` acRange c2
                -- NOTE: is it possible that we get a wrong argument order when doing `apply` because of this concatenation?
```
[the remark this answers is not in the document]
echatav: removed

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
echatav: removed commented code

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: Discussed it in #154

src/ZkFold/Symbolic/Data/FieldElement.hs
TurtlePU: Discussed in #154

tests/Tests/ArithmeticCircuit.hs
TurtlePU: Discussed in #154

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
vlasin: No, it's not needed. However, some minor changes will be required in the `toPlonkRelation` function.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
vlasin: Yes, we could have multiple constraints with the same constants that depend on different subsets of input variables. This would give the same new variable every time if I am not mistaken. So, it must be fixed in this PR.

TurtlePU: As a hotfix, we can require `ToConstant (Rep i) Natural` before we migrate to the proper solution

src/ZkFold/Symbolic/Cardano/Types/Address.hs
TurtlePU: Written down in #113

src/ZkFold/Symbolic/Cardano/Types/Output.hs
TurtlePU: Written down in #113

TurtlePU (approved):

TurtlePU (dismissed):

vlasin:
Pinging @vks4git to check the Protostar edits. No issues from my perspective :+1:

vlasin (dismissed):

TurtlePU (dismissed):
LGTM2! Only conflicts in examples/ introduced by #187 need to be fixed

vks4git (dismissed):

src/ZkFold/Base/Protocol/ARK/Plonk/Relation.hs
TurtlePU: @echatav, if all tests pass, you can resolve this conversation

echatav: @TurtlePU Neither my laptop nor CI are adequate to check if all tests pass :-/

TurtlePU: @echatav it's ok, I'll run them myself then once merge conflicts are resolved

vlasin: I just want to point out that there is a very unlikely scenario when the quoted line results in "index overflow". This is something we want to solve in the var generation PR.

vlasin: Sorry, my bad! We changed this conversion algorithm some time ago, so it should not cause any problems.

echatav: I'd like to have a discussion to understand var generation better.

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -46,31 +46,31 @@ data Plonk (n :: Natural) (l :: Natural) curve1 curve2 transcript = Plonk {
         k1    :: ScalarField curve1,
         k2    :: ScalarField curve1,
         iPub  :: Vector l Natural,
-        ac    :: ArithmeticCircuit (ScalarField curve1) Par1,
+        ac    :: ArithmeticCircuit (ScalarField curve1) (Vector l) Par1,
```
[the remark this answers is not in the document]
TurtlePU: Looks like "circuit inputs" and "public inputs" are actually two different concepts, so I'm lost on what the second concept should mean

tests/Tests/Arithmetization/Test4.hs
```diff
@@ -37,23 +38,25 @@ lockedByTxId targetValue inputValue = inputValue == fromConstant targetValue
...
-    let Bool ac = compile @F (lockedByTxId @F @(ArithmeticCircuit F) targetValue) :: Bool (ArithmeticCircuit F)
+    let Bool ac = compile @F (lockedByTxId @F @(ArithmeticCircuit F (V.Vector 1)) targetValue) :: Bool (ArithmeticCircuit F (V.Vector 1))
 
         (omega, k1, k2) = getParams 32
-        witnessInputs   = fromList [(1, targetValue), (unPar1 $ acOutput ac, 1)]
-        indexOutputBool = V.singleton $ unPar1 $ acOutput ac
+        witnessInputs = V.singleton targetValue
+        indexOutputBool = V.singleton $ case unPar1 $ acOutput ac of
+          NewVar ix -> ix + 1
+          InVar _   -> 1
```
zlonast, line 59: Looks scary, I hope the user will call some library function

src/ZkFold/Base/Protocol/ARK/Plonk.hs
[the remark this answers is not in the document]
zlonast: (l :: Natural) here is the number of public inputs to plonk. Plonk has no outputs as such. I'm not sure what we should do here.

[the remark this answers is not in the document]
TurtlePU: So from discussion with Ilya I gathered that "public inputs" of Plonk protocol is actually some arbitrary subset of input & output variables. Consider adding an additional parameter `i` to Plonk-related datatypes with `ac :: ArithmeticCircuit (ScalarField curve1) i Par1` and `iPub :: Vector l (Var i)`.
Maybe even `iPub :: (i Bool, o Bool)` if `l` can be safely removed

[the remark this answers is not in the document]
TurtlePU: > Plonk has no outputs

So how is the output to which `Par1` points to is used? Is it left unused and has to be constrained via including it to public inputs?

TurtlePU:
I suggest we first fix the bug where we mixed (by mistake) circuit inputs and public inputs and finally merge this PR

For the future, I suggest we assume *the outputs of a circuit are what Plonk calls public inputs*. This makes sense because this is, like, the output of a computation we can check against. This also simplifies the implementation of Plonk as this allows to discard `iPub` parameter and reuse output manipulation primitives from `ArithmeticCircuit` API.

TurtlePU:
Honestly I'd refactor Plonk modules a bit...

TurtlePU (approved):
FINALLY

vlasin (approved):

TurtlePU:
🎆

PR #215, opened 2024-08-12
Vks4git/protostar

vks4git:


src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -162,7 +162,7 @@ class MultiplicativeMonoid b => Scale b a where
 
 instance MultiplicativeMonoid a => Scale a a
 
-instance {-# OVERLAPPABLE #-} (Scale b a, Functor f) => Scale b (f a) where
+instance {-# OVERLAPPABLE #-} (Scale b a, Functor f, MultiplicativeMonoid b, fa ~ f a) => Scale b fa where
```
TurtlePU, line 165: This instance was annoying me as well, I was going to submit a PR which removed this instance in addition to some other improvements to numeric hierarchy

src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -53,6 +54,9 @@ variables :: forall c v .
     Poly c v Natural -> Set v
 variables = runSources . evalPolynomial evalMonomial (Sources @c . singleton)
 
+mapVars :: Variable i2 => (i1 -> i2) -> Poly c i1 j -> Poly c i2 j
```
TurtlePU, line 57: FINALLY

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -59,9 +62,20 @@ newtype Poly c = P (V.Vector c)
...
+-- | A polynomial of form cx^d
+--
+monomial :: (Ring c, Eq c) => Natural -> c -> Poly c
+monomial d c = toPoly $ V.fromList (replicate d zero P.<> [c])
+
 fromPoly :: Poly c -> V.Vector c
 fromPoly (P cs) = cs
 
+evalPoly :: Ring c => Poly c -> c -> c
+evalPoly (P cs) x = sum $ V.zipWith (*) cs $ fmap (x^) (V.generate (V.length cs) (fromIntegral @_ @Natural))
```
TurtlePU, line 77: Urges to introduce some common interface to different kinds of polynomials, will add a proposal about this later

src/ZkFold/Base/Data/Vector.hs
```diff
@@ -53,7 +59,10 @@ unsafeToVector :: forall size a . [a] -> Vector size a
 unsafeToVector = Vector
 
 generate :: forall size a . KnownNat size => (Natural -> a) -> Vector size a
-generate f = Vector $ f <$> [0 .. value @size -! 1]
+generate f = Vector $
+    case value @size of
+      0 -> [] -- avoid arithmetic underflow
```
TurtlePU, line 64: :heart:

src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
```diff
@@ -1,116 +1,141 @@
...
+data KeyScale f = KeyScale f f
+    deriving (P.Show, Generic, NFData)
+
+-- | Class describing types which can form a polynomial linear combination:
+-- linearCombination a1 a2 -> a1 * X + a2
+--
+class LinearCombination a b | a -> b where
+    linearCombination :: a -> a -> b
+
+-- | Same as above, but with a coefficient known at compile time
```
TurtlePU, line 58: Maybe vice-versa, here coefficient is known only at runtime?

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
```diff
@@ -0,0 +1,135 @@
...
+-- | A type for recurcive computations.
+-- @circuit@ is an Arithmetic circuit with @n@ inputs and @n@ outputs applied to itself (i.e. outputs are fed as inputs at the next iteration) @iterations@ times.
+--
+data RecursiveCircuit n a
+    = RecursiveCircuit
+        { iterations :: Natural
+        , circuit    :: ArithmeticCircuit a (Vector n) (Vector n)
+        } deriving (Generic, NFData)
+
+instance (Ring f, P.Eq f) => Iso (PU.Poly f) f where
```
TurtlePU, line 53: Is this really an isomorphism? Let $p(x) = x+1$, then `from (from p) = 1`

src/ZkFold/Base/Protocol/Protostar/Commit.hs
```diff
@@ -0,0 +1,71 @@
...
+class PedersonSetup a where
+    pedersonGH :: (a, a)
+
+instance PedersonSetup (Point BLS12_381_G1) where
+    pedersonGH = (g, h)
+        where
+            -- Random points on BLS12_381_G1
+            -- The only requirement for them is so that nobody knows discrete logarithm of g base h
+            -- Keeping these numbers open seems safe as there is no known efficient algorithm to calculate discrete logarithm
+            -- TODO: Consider choosing these elements randomly each time instead of hardcoding them
```
TurtlePU, line 39: There is no known efficient algorithm, but as long as parameters are fixed, one can ran an inefficient algorithm long enough, so yeah, maybe we will choose them randomly soon

src/ZkFold/Base/Protocol/Protostar/Commit.hs
```diff
@@ -0,0 +1,71 @@
...
+    , BinaryExpansion b
+    , Bits a ~ [a]
+    , Bits b ~ [b]
+    ) => HomomorphicCommit a b (Point BLS12_381_G1) where
+    hcommit r b = let (g, h) = pedersonGH @(Point BLS12_381_G1)
+                   in pointMul b g + pointMul r h
+
+
+-- Pedersen commitment scheme for lists extending the homomorphism to elementwise sums:
+-- (hcommit ck l1) + (hcommit ck l2) = hcommit ck (zipWith (+) l1 l2), provided l1 and l2 have the same length.
```
TurtlePU, line 65: > provided `l1` and `l2` have the same length
Maybe it is better to specialize `t` to `Vector n`, then? Or is `n` unknown at compile-time

src/ZkFold/Base/Protocol/Protostar/Oracle.hs
```diff
@@ -1,6 +1,72 @@
...
+instance (AdditiveMonoid b, RandomOracle a b) => RandomOracle [a] b where
+    oracle as = sum $ oracle <$> as
+
+instance {-# OVERLAPPABLE #-} (Generic a, RandomOracle' (Rep a) b) => RandomOracle a b where
+
+class RandomOracle' f b where
+    oracle' :: f a -> b
+
+instance RandomOracle' V1 b where
+    oracle' = P.undefined -- No oracle for uninhabited types
```
TurtlePU, line 44: Can this instance be removed, then? This will lead to better compilation errors, too

src/ZkFold/Base/Protocol/Protostar/Oracle.hs
```diff
@@ -1,6 +1,72 @@
...
+
+instance RandomOracle' U1 b where
+    oracle' = P.undefined -- No useful information can be extracted from U1.
```
TurtlePU, line 47: Can this instance be removed, then? This will lead to better compilation errors, too

tests/Tests/Protostar.hs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+instance P.Show a => P.Show (RecursiveFunction n c a) where
+    show RecursiveFunction{..} = P.unlines [P.show rIterations, P.show rInitial]
+
+instance
+    ( KnownNat n
+    , Arbitrary a
+    , Symbolic c
+    , MultiplicativeSemigroup (FieldElement c)
```
TurtlePU, line 41: This should be inferrable from `Symbolic c` constraint

zkfold-base.cabal
```diff
@@ -342,4 +347,26 @@ benchmark compiler-benchmark
       tasty-bench,
       tasty-golden,
       zkfold-base,
-      zkfold-symbolic-examples
+      zkfold-symbolic-examples,
+      time,
```
TurtlePU, line 351: These dependencies are not needed in `compiler-benchmark`

src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -0,0 +1,179 @@
...
+    => P.Eq c
+    => AdditiveGroup c
+    => Scale a c
+    => Scale a a
+    => KnownNat n
+    => RandomOracle a a
+    => RandomOracle c a
+    => RandomOracle [c] a
+    => HomomorphicCommit a [a] c
+    => HomomorphicCommit a [Map Natural a] c
```
TurtlePU, line 128: These look like references to concrete instances, maybe it is better to inline them then?

src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -0,0 +1,179 @@
...
+    => x ~ ArithmeticCircuit a (Vector n)
+    => P.Eq c
+    => Scale a c
+    => Scale a a
+    => AdditiveGroup c
+    => RandomOracle a a
+    => RandomOracle c a
+    => RandomOracle [c] a
+    => HomomorphicCommit a [a] c
+    => HomomorphicCommit a [Map Natural a] c
```
TurtlePU, line 83: These look like references to concrete instances, maybe it is better to inline them then?

src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -0,0 +1,179 @@
...
+    , lastAccumulator :: Accumulator (Vector n a) a c (Map Natural a)
+    , verifierOutput  :: P.Bool
+    , deciderOutput   :: P.Bool
+    } deriving (P.Show, Generic, NFData)
+
+type FS_CM n c a = FiatShamir a (CommitOpen a c (RecursiveCircuit n a))
+
+transform
+    :: forall n c a
+    .  HomomorphicCommit a [Map Natural a] c
```
TurtlePU, line 47: These look like references to concrete instances, maybe it is better to inline them then?

TurtlePU (commented):
:fire: :fire: :fire: 

Excellent implementation with good additions along the way, but I have my doubts on the use of typeclasses in Protostar implementation

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
vks4git: No, of course it isn't. I was only trying to make it work and used an available class. I'll create another class

src/ZkFold/Base/Protocol/Protostar/Commit.hs
vks4git: n is unknown. Foldable types here can be witness Maps and constraint systems

tests/Tests/Protostar.hs
vks4git: Yeah, there are a lot of cases where GHC does not flag redundant constraints. In some cases where constraint relations were not obvious I even had to remove constraints one-by-one and observe if it would lead to errors

zkfold-base.cabal
vks4git: Probably rebase artifacts

src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
```diff
@@ -1,116 +1,141 @@
...
-  commit (FiatShamir (CommitOpen cm _) _) = cm
-
-  prover sps@(FiatShamir (CommitOpen cm _) _) acc (InstanceProofPair pubi (NARKProof pi_x pi_w)) =
-      (Accumulator (AccumulatorInstance pi'' ci'' ri'' eCapital' mu') mi'', eCapital_j)
+    , AdditiveSemigroup m
+    , Ring f
+    , Scale f c
+    , Scale f m
+    , Input f a ~ i
+    , LinearCombination (ProverMessage f a) (ProverMessage (PU.Poly f) a)
```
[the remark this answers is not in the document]
vks4git: Nope, the algorithm should be able to form a polynomial from any prover messages. 
See https://eprint.iacr.org/2023/620.pdf , page 21, figure 3

src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
TurtlePU: Written down in https://github.com/zkFold/zkfold-base/issues/154#issuecomment-2337702244

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
TurtlePU: Is typeclass needed for this?

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
```diff
@@ -0,0 +1,135 @@
...
+--}
+
+
+-- | A type for recurcive computations.
+-- @circuit@ is an Arithmetic circuit with @n@ inputs and @n@ outputs applied to itself (i.e. outputs are fed as inputs at the next iteration) @iterations@ times.
+--
+data RecursiveCircuit n a
+    = RecursiveCircuit
+        { iterations :: Natural
+        , circuit    :: ArithmeticCircuit a (Vector n) (Vector n)
```
vlasin, line 50: This circuit is for the following recursion:
$x_{i+1} = F(x_i)$.
I believe Protostar supports a more general case:
$exists w_i:~ x_{i+1} = F(x_i, w_i)$

`pi` from the paper are "public inputs" to the circuit, `w` are private inputs. If I am correct, a recursion circuit does not need to have outputs at all, like in our general `ArithmeticCircuit` definition.

src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -0,0 +1,179 @@
...
+
+        nark@(InstanceProofPair _ narkProof) = instanceProof ck rc i
+        (newAcc, accProof) = Acc.prover fs ck acc nark
+
+        alpha :: a
+        alpha = oracle (acc^.x, i, narkCommits narkProof, accProof)
+
+        alphaPows = sum $ P.take (P.length accProof) $ (alpha ^) <$> [1 :: Natural ..]
+
+        verifierAccepts = Acc.verifier @_ @_ @_ @(Map Natural a) @(FS_CM n c a) i (narkCommits narkProof) (acc^.x) (newAcc^.x) accProof
```
vlasin, line 177: The verifier should always accept the proof when it is constructed with the `prover` algorithm. These checks are redundant here.

src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -0,0 +1,179 @@
...
+    => RandomOracle c a
+    => RandomOracle [c] a
+    => HomomorphicCommit a [a] c
+    => HomomorphicCommit a [Map Natural a] c
+    => KnownNat n
+    => (Vector n (FieldElement x) -> Vector n (FieldElement x))  -- ^ An arithmetisable function to be applied recursively
+    -> Natural                             -- ^ The number of iterations to perform
+    -> SPS.Input a (RecursiveCircuit n a)  -- ^ Input for the first iteration
+    -> FoldResult n c a
+fold f iter i = foldN iter ck rc i [] initialAccumulator (Acc.KeyScale one one)
```
vlasin, line 89: In order to make it a "Symbolic fold", it should return a `SymbolicData` type. In this particular case of arithmetisable function, the result should be `Vector n (FieldElement x)`. See also my comment about the `RecursiveCircuit` type.

In the "circuit" context, the IVC prover algorithm will make its way into witness generation, while the IVC verifier (arithmetised decider + arithmetised one step) will be added to the overall circuit.

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
vks4git: The problem is, `algebraicMap` from `SpecialSoundProtocol` is polymorphic. Depending on the parameter `f`, it can either construct a system of equations if `f` is a polynomial or just compute a vector if `f` is a field element. This behaviour is required in the algorithm in `AccumulatorScheme`. Compiler doesn't like that `algebraicMap` is called with different type parameters, that's why I gave it a hint that the types can be converted into each other. Let me think if I can do it more nicely.

src/ZkFold/Base/Protocol/Protostar/ArithmeticCircuit.hs
```diff
@@ -0,0 +1,118 @@
...
+  , Scale a f
+  , MultiplicativeMonoid f
+  , Exponent f Natural
+  , AdditiveMonoid f
+  ) => SPS.AlgebraicMap f (ArithmeticCircuit a (Vector n) o) where
+
+    type MapInput f (ArithmeticCircuit a (Vector n) o) = Vector n f
+    type MapMessage f (ArithmeticCircuit a (Vector n) o) = Map ByteString f
+
+    -- We can use the polynomial system from the xircuit as a base for V_sps.
```
vlasin, line 75: Type `circuit`

src/ZkFold/Symbolic/Algorithms/Hash/MiMC/Constants.hs
```diff
@@ -1,21 +1,235 @@
...
-import           Data.Maybe                      (fromJust)
 import           Prelude
 
 import           ZkFold.Base.Algebra.Basic.Class (FromConstant (..))
-import           ZkFold.Base.Data.ByteString
 
-mimcSeed :: LittleEndian
-mimcSeed = 42
+cs :: [Integer]
+cs = [42,
```
vlasin, line 8: Why are we making this change?

vks4git: For speed. The list of constants was generated every time mimcHash was called. It took too long. This list is just the hardcoded result of the constants generator

src/ZkFold/Base/Algebra/EllipticCurve/Class.hs
```diff
@@ -40,6 +47,14 @@ instance (EllipticCurve curve, Eq (BaseField curve)) => Eq (Point curve) where
 instance EllipticCurve curve => AdditiveSemigroup (Point curve) where
     (+) = add
 
+instance {-# OVERLAPPABLE #-}
+    ( EllipticCurve curve
+    , Eq s
+    , BinaryExpansion s
+    , Bits s ~ [s]
+    ) => Scale s (Point curve) where
```
TurtlePU, line 55: In principle, this instance does the same thing that `Scale Natural` does. Maybe we can replace it with a single
```haskell
instance (EllipticCurve curve, ToConstant s, Const s ~ Natural) => Scale s (Point curve) where
  scale = natScale . toConstant
```
?

TurtlePU: (Though this is not really lawful because we also need `s` to have appropriate order in case it is finite...)

TurtlePU: Though this is a topic for another PR as this would invlove changing some of the elliptic curve code, I guess

src/ZkFold/Symbolic/Algorithms/Hash/MiMC/Constants.hs
TurtlePU: Can it be a `take 218 $ map getI $ iterate hash $ toByteString mimcSeed` at least? Maybe GHC would compute it only once this way

TurtlePU:
Ready to approve, too, but the fact that proving protocol has to use Symbolic DSL API bothers me, will think about it when I get back from my vacation

vlasin:
> Ready to approve, too, but the fact that proving protocol has to use Symbolic DSL API bothers me, will think about it when I get back from my vacation

We take one circuit (recursion step) and construct another one (full recursion). But maybe there is a way to decouple it like you suggested with Plonk.

TurtlePU:
> We take one circuit (recursion step) and construct another one (full recursion).

Sounds like something to be exposed via Symbolic API to end users to be used in recursive datatype definitions and recursive functions! Will look into it once I get back from vacation, sounds very interesting.

vlasin (approved):

PR #216, opened 2024-08-12
Symbolic UInt

hovanja2011:


src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-    , from ~ NumberOfRegisters a n r
-    , to ~ NumberOfRegisters a k r
-    ) => Extend (UInt n r (ArithmeticCircuit a)) (UInt k r (ArithmeticCircuit a)) where
-    extend (UInt ac) = UInt (circuitF solve)
+    , Finite (BaseField с)
+    , ToConstant (Vector (NumberOfRegisters (BaseField c) n r) (BaseField c)) Natural
+    ,FromConstant Natural (Vector (NumberOfRegisters (BaseField c) k r) (BaseField c))
+    -- , BaseField c ~ BaseField c
+    ) => Extend (UInt n r c) (UInt k r c) where
+    extend (UInt x) =  UInt $ symbolicF x (\l -> fromConstant (toConstant l :: Natural)) solve
```
vlasin, line 163: Doing bit expansion to extend the number seems not optimal to me. Is there a better solution?

src/ZkFold/Symbolic/Cardano/Types/Value.hs
```diff
@@ -30,7 +30,7 @@ deriving instance
     , KnownNat (TypeSize context (SingleAsset context))
     ) => SymbolicData context (Value n context)
 
-instance (FromConstant Natural (UInt 64 Auto context), MultiplicativeSemigroup (UInt 64 Auto context))
+instance (Symbolic context, FromConstant Natural (UInt 64 Auto context), MultiplicativeSemigroup (UInt 64 Auto context))
```
TurtlePU, line 33: Maybe with `Symbolic context` constraint, we can get rid of `FromConstant Natural ...` and `MultiplicativeSemigroup ...` constraints?

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -147,6 +147,10 @@ class BitState c n b where
     isUnset :: c n b -> Natural -> Bool b
 
 
+instance ToConstant (ByteString n (ArithmeticCircuit a)) Natural where
+    toConstant (ByteString ac) = Haskell.foldl (\y p -> toConstant p + base * y) 0 (acOutput ac)
```
TurtlePU, line 151: This does the wrong thing because `acOutput` contains indices of variables so it doesn't make sense to me to add them

TurtlePU: Also in general there cannot be a `ToConstant (smth of arithmetic circuits) Natural` instance because arithmetic circuit might have unfed inputs (this would become apparent after we merge #214).

src/ZkFold/Symbolic/Data/Ed25519.hs
```diff
@@ -30,17 +31,17 @@ import           ZkFold.Symbolic.Data.Eq
 import           ZkFold.Symbolic.Data.UInt
 import           ZkFold.Symbolic.Interpreter
 
-zpToEd :: Finite (Zp p) => Point (Ed25519 (Interpreter (Zp p))) -> Point (Ed25519 Void)
+zpToEd :: (Symbolic (Interpreter (Zp p)), FromConstant (Interpreter (Zp p) Par1) Natural ) => Point (Ed25519 (Interpreter (Zp p))) -> Point (Ed25519 Void)
```
TurtlePU, line 34: I feel like we do not have a `FromConstant (Interpreter ...) Natural` instance (and interpreter is not constant, so it would be strange to call it that)

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -15,89 +15,64 @@ module ZkFold.Symbolic.Data.UInt (
...
-instance (KnownNat n, Finite (Zp p), KnownRegisterSize r) => FromConstant Natural (UInt n r (Interpreter (Zp p))) where
+instance (Symbolic c, KnownNat n, KnownRegisterSize r, FromConstant a Natural) => FromConstant a (UInt n r c) where
     fromConstant c =
-        let (lo, hi, _) = cast @(Zp p) @n @r . (`Haskell.mod` (2 ^ getNatural @n)) $ c
-         in UInt $ Interpreter $ V.unsafeToVector $ (fromConstant <$> lo) <> [fromConstant hi]
+        let c' = toConstant . (fromConstant :: a -> Natural) $ c
+            (lo, hi, _) = cast @(BaseField c) @n @r . (`Haskell.mod` (2 ^ getNatural @n)) $ c'
+        in UInt . embed @c $ V.unsafeToVector $ (fromConstant <$> lo) <> [fromConstant hi]
 
-instance (KnownNat n, Finite (Zp p), KnownRegisterSize r) => FromConstant Integer (UInt n r (Interpreter (Zp p))) where
```
TurtlePU, line 72: `FromConstant Integer` instance is lost

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -15,89 +15,64 @@ module ZkFold.Symbolic.Data.UInt (
...
-instance
-    ( FromConstant Natural a
-    , Arithmetic a
-    , KnownNat n
-    , KnownRegisterSize r
-    ) => FromConstant Integer (UInt n r (ArithmeticCircuit a)) where
-    fromConstant = fromConstant . naturalFromInteger . (`Haskell.mod` (2 ^ getNatural @n))
-
-instance (FromConstant Natural (UInt n r b), KnownNat n, MultiplicativeSemigroup (UInt n r b)) => Scale Natural (UInt n r b)
+instance (Symbolic c, FromConstant a (UInt n r c), MultiplicativeSemigroup (UInt n r c), MultiplicativeMonoid a) => Scale a (UInt n r c)
```
TurtlePU, line 70: `MultiplicativeSemigroup (UInt n r c)` instance should be redundant

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
 -- If the algorithm is only used to find @gcd(a, b)@ (i.e. @s@ and @t@ will be discarded), @a@ and @b@ can be arbitrary integers.
 --
 eea
-    :: forall n b r
-    .  Symbolic b
-    => EuclideanDomain (UInt n r b)
+    :: forall n c r
+    .  Symbolic c
+    => EuclideanDomain (UInt n r c)
```
TurtlePU, line 107: All constraints except for `Symbolic` and `KnownNat` have to be redundant

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
           | iteration Haskell.== iterations = (oldS, oldT, oldR)
-          | otherwise = bool @(Bool b) rec (if Haskell.even iteration then b - oldS else oldS, if Haskell.odd iteration then a - oldT else oldT, oldR) (r == zero)
+          | otherwise = bool @(Bool c) rec (if Haskell.even iteration then b - oldS else oldS, if Haskell.odd iteration then a - oldT else oldT, oldR) (r == zero)
             where
                 quotient = oldR `div` r
 
                 rec = eea' (iteration + 1) r (oldR - quotient * r) s (quotient * s + oldS) t (quotient * t + oldT)
 
 --------------------------------------------------------------------------------
+instance (Symbolic c, KnownNat n, KnownRegisterSize r, Haskell.Num b, FromConstant (c Par1) Natural) => ToConstant (UInt n r c) b where
```
TurtlePU, line 128: It's ok to keep the `ToConstant (UInt n r (Interpreter a)) Natural` instance for now, as we do not have the means to properly generalize it yet

Also we do not have a `FromConstant (c Par1) Natural` instance

src/ZkFold/Symbolic/Data/UInt.hs
TurtlePU: In case of `r ~ Auto` -- no, unfortunately, as changing the `n` might change the register size.
However in case of `r ~ Fixed rs`, we can pad with the necessary amount of zeros, for sure.

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-            fullSub :: MonadBlueprint i a m => i -> i -> i -> m (i, i)
+            fullSub :: MonadCircuit i (BaseField c) m => i -> i -> i -> m (i, i)
             fullSub xk yk b = do
                 k <- newAssigned (\v -> v xk - v yk)
                 s <- newAssigned (\v -> v k + v b + fromConstant t)
-                splitExpansion (registerSize @a @n @r) 1 s
+                splitExpansion (registerSize @(BaseField c) @n @r) 1 s
 
-    strictMul (UInt x) (UInt y) = UInt (circuitF $ V.unsafeToVector <$> solve)
+    strictMul (UInt x) (UInt y) = UInt $ symbolic2F x y (\u v -> V.unsafeToVector $ V.fromVector u * V.fromVector v) solve
```
TurtlePU, line 477: Pointwise list addition doesn't do what is needed here, I would write a helper function `Arithmetic a => Vector (NumberOfBits ...) a -> Natural` which does the same what `ToConstant (UInt n r (Interpreter a)) Natural` does

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
+                    Haskell.zipWith (fullAdder $ registerSize @(BaseField c) @n @r) midx midy
                 k <- fullAdded z w c
-                _ <- expansion (highRegisterSize @a @n @r) k
-                return (zs <> [k])
+                (ks, _) <- splitExpansion (highRegisterSize @(BaseField c) @n @r) 1 k
+                return $ V.unsafeToVector (zs ++ [ks])
 
 
-    strictSub (UInt x) (UInt y) = UInt (circuitF $ V.unsafeToVector <$> solve)
+    strictSub (UInt x) (UInt y) = UInt $ symbolic2F x y (\u v -> V.unsafeToVector $ V.fromVector u - V.fromVector v) solve
```
TurtlePU, line 440: Same here, poitwise negation is not what is needed here

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-    , KnownRegisterSize r
-    , (NumberOfRegisters a n r - 1) + 1 ~ NumberOfRegisters a n r
-    ) => Ring (UInt n r (ArithmeticCircuit a))
+    , (NumberOfRegisters (BaseField c) n r - 1) + 1 ~ NumberOfRegisters (BaseField c) n r
+    , FromConstant Integer Natural
+    , ToConstant (Vector (NumberOfRegisters (BaseField c) n r) (BaseField c)) Natural
+    , FromConstant Natural (Vector (NumberOfRegisters (BaseField c) n r) (BaseField c))
+    ) => Ring (UInt n r c)
 
 deriving via (Structural (UInt n rs (ArithmeticCircuit a)))
```
TurtlePU, line 388: Can be generalized to arbitrary `Symbolic c` instead of `c ~ ArithmeticCircuit a`

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-instance
-    ( Arithmetic a
-    , KnownNat n
-    , KnownNat (NumberOfRegisters a n r)
-    , KnownNat (NumberOfRegisters a n r - 1)
-    , KnownRegisterSize r
-    , (NumberOfRegisters a n r - 1) + 1 ~ NumberOfRegisters a n r
-    ) => Ring (UInt n r (ArithmeticCircuit a))
+    , (NumberOfRegisters (BaseField c) n r - 1) + 1 ~ NumberOfRegisters (BaseField c) n r
+    , FromConstant Integer Natural
```
TurtlePU, line 383: `FromConstant Integer Natural` cannot be a lawful instance so we do not have it

src/ZkFold/Symbolic/Data/UInt.hs
vlasin: Ok, agree!

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-
-instance
-    ( Arithmetic a
-    , KnownNat n
-    , KnownNat (NumberOfRegisters a n r)
-    , KnownNat (NumberOfRegisters a n r - 1)
-    , KnownRegisterSize r
-    , (NumberOfRegisters a n r - 1) + 1 ~ NumberOfRegisters a n r
-    ) => Ring (UInt n r (ArithmeticCircuit a))
+    , (NumberOfRegisters (BaseField c) n r - 1) + 1 ~ NumberOfRegisters (BaseField c) n r
```
TurtlePU, line 382: This constraint is unfortunate to have, where does it come from and can we get rid of it?

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
+    , FromConstant Integer Natural
+    , ToConstant (Vector (NumberOfRegisters (BaseField c) n r) (BaseField c)) Natural
```
TurtlePU, line 384: `ToConstant a Natural` is a wrong instance to have for any `a` except `Natural`, where is it used?

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
+    , FromConstant Natural (Vector (NumberOfRegisters (BaseField c) n r) (BaseField c))
```
TurtlePU, line 385: It's strange that we need this instance

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-
-                negateN :: MonadBlueprint i a m => Natural -> i -> i -> m (i, i)
-                negateN n i b = do
-                    r <- newAssigned (\v -> fromConstant n - v i + v b)
-                    splitExpansion (registerSize @a @n @r) 1 r
-
-
-instance (Arithmetic a, KnownNat n, KnownRegisterSize rs, r ~ NumberOfRegisters a n rs) => MultiplicativeSemigroup (UInt n rs (ArithmeticCircuit a)) where
-    UInt x * UInt y = UInt (circuitF $ V.unsafeToVector <$> solve)
+    UInt x - UInt y = UInt $ symbolic2F x y (\u v -> V.unsafeToVector $ V.fromVector u + V.fromVector v) solve
```
TurtlePU, line 278: Once again, pure implementation must come from the old instance with `Interpreter` context

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
+    zero = fromConstant (0:: Natural)
 
 instance
-    ( Arithmetic a
+    (Symbolic c
     , KnownNat n
     , KnownRegisterSize r
-    , KnownNat (NumberOfRegisters a n r)
-    ) => AdditiveGroup (UInt n r (ArithmeticCircuit a)) where
+    , FromConstant Integer Natural
```
TurtlePU, line 274: These `FromConstant`/`ToConstant` constraints are wrong to have

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -127,522 +102,417 @@ cast n =
...
-    , AdditiveGroup (UInt n rs b)
-    , Semiring (UInt n rs b)
-    , MultiplicativeMonoid (UInt n rs b)
-    , FromConstant Natural (UInt n rs b)
-    , BitState ByteString n b
-    , Iso (ByteString n b) (UInt n rs b)
-    , Eq (Bool b) (UInt n rs b)
+    , r ~ NumberOfRegisters (BaseField c) n rs
+    , NFData (c (Vector r))
+    , Ord (Bool c) (UInt n rs c)
```
TurtlePU, line 200: All the constraints on `UInt` except for `Iso` have to be inferred from `Symbolic c`

TurtlePU:
Cheat-sheet of instances I advise to have after this PR:
* derived:
  *  `Generic`, `SymbolicData`
  * with appropriate superclass constraints -- `Show`, `Haskell.Eq`, `NFData`
  * via `Structural` -- `Symbolic c => Eq (Bool c)`
* implemented to work in a generic `Symbolic c` context without any extra constraints:
  * `Ord (Bool c)`, `StrictConv Natural`, `StrictConv (FieldElement c)`, `StrictNum`
  * `Ring` with all superclasses (`Scale Natural` and `Scale Integer` can have default impls)
  * `Arbitrary`, keeping the current implementation of just promoting constant value to `c`
  * `Extend`, `Shrink`
* implemented to work in an `Interpreter` context:
  * `ToConstant (UInt n r (Interpreter a)) Natural`
* kept to be refactored later:
  *  `Iso` with `ByteString`
  * `EuclideanDomain (UInt n rs c)`

TurtlePU:
Slightly off-topic for this PR, just a thought on entanglement between `ByteString` and `UInt` modules
If I remember correctly, @vks4git was proposing to implement bitwise operations for UInt directly instead of converting between UInt and ByteString
And now I'm thinking that maybe he was right because with lookup arguments, we can have more efficient bitwise operations without breaking things into bits so we can avoid extra constraints from reconverting between representations

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -652,10 +541,20 @@ instance (Arithmetic a, KnownNat n, KnownRegisterSize r) => StrictNum (UInt n r
...
+fullAdder :: (Arithmetic a, MonadCircuit i a m) => Natural -> i -> i -> i -> m (i, i)
 fullAdder r xk yk c = fullAdded xk yk c >>= splitExpansion r 1
 
-fullAdded :: MonadBlueprint i a m => i -> i -> i -> m i
+fullAdded :: MonadCircuit i a m => i -> i -> i -> m i
 fullAdded i j c = do
     k <- newAssigned (\v -> v i + v j)
     newAssigned (\v -> v k + v c)
+
+naturalToVector :: forall c n r . (Symbolic c, KnownNat n, KnownRegisterSize r) => Natural -> Vector (NumberOfRegisters (BaseField c) n r) (BaseField c)
```
TurtlePU, line 552: We can call to this function from `FromConstant` to avoid repetition

src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -652,10 +541,20 @@ instance (Arithmetic a, KnownNat n, KnownRegisterSize r) => StrictNum (UInt n r
...
+naturalToVector c = let (lo, hi, _) = cast @(BaseField c) @n @r . (`Haskell.mod` (2 ^ getNatural @n)) $ c
+    in V.unsafeToVector $ (fromConstant <$> lo) <> [fromConstant hi]
+
+
+vectorToNatural :: (ToConstant a Natural) => Vector n a -> Natural -> Natural
```
TurtlePU, line 557: We can call to this function from `toConstant` to avoid repetition

TurtlePU:
Nice! Only `Ord` instance left practically

TurtlePU (dismissed):

TurtlePU (approved):

PR #218, opened 2024-08-15
feat: add proof parametrization and benchmark

diS3e:


src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -39,29 +40,29 @@ import           ZkFold.Symbolic.MonadCircuit               (Arithmetic)
     Additionally, we don't want this library to depend on Cardano libraries.
 -}
 
-data Plonk (n :: Natural) (l :: Natural) curve1 curve2 transcript = Plonk {
+data Plonk (n :: Natural) (l :: Natural) curve1 curve2 transcript core =
+    CoreFunction curve1 core =>
```
zlonast, line 44: Are you sure we don't want to add core only to `NonInteractiveProof`? 🤔

diS3e: Yes, you are right. Here I can remove the core from the definition and leave it only in the constraint for the NonInteractiveProof instance

vlasin:
I have two suggestions:

1. Rename `com'` to `msm` for clarity.
2. Move the benchmark to the `zkfold-prover` repository, because there you can compare two implementations in one benchmark.

zlonast (approved):

vlasin (approved):

PR #220, opened 2024-08-17
Binary fix + Plonk protocol update

vlasin:
- [x] Fixed `Binary` for the compressed and uncompressed points on BLS12_381 "G1" and "G2" curves.
- [x] Changed `Ord` for field extensions to a more natural one.
- [x] Added tests for the fixed instances.
- [x] In the Plonk protocol, we now use compressed points for shorter transcripts.

NOTE: The test for `Binary` instance for the BLS curve points is incomplete. The instance allows us to `get` uncompressed `Point` from the short bytestring and vice versa. However, we don't check the correctness in these cases. Do we even need such flexibility? We can compress/uncompress already restored points.


src/ZkFold/Base/Algebra/EllipticCurve/Class.hs
```diff
@@ -123,6 +123,23 @@ class EllipticCurve curve => StandardEllipticCurve curve where
...
+    show (PointCompressed x bigY) = "(" ++ show x ++ ", " ++ show bigY ++ ")"
+
+instance Eq (BaseField curve) => Eq (PointCompressed curve) where
+    InfCompressed == InfCompressed                       = True
+    PointCompressed x1 bigY1 == PointCompressed x2 bigY2 = x1 == x2 && bigY1 == bigY2
+    _ == _                                               = False
+
+instance (Arbitrary (Point curve), AdditiveGroup (BaseField curve), Ord (BaseField curve)
+        ) => Arbitrary (PointCompressed curve) where
+    arbitrary = do
```
TurtlePU, line 137: This can be done as `compress <$> arbitrary`

src/ZkFold/Base/Algebra/EllipticCurve/Class.hs
```diff
@@ -123,6 +123,23 @@ class EllipticCurve curve => StandardEllipticCurve curve where
 
 data PointCompressed curve = PointCompressed (BaseField curve) Bool | InfCompressed
 
+instance Show (BaseField curve) => Show (PointCompressed curve) where
+    show InfCompressed            = "InfCompressed"
+    show (PointCompressed x bigY) = "(" ++ show x ++ ", " ++ show bigY ++ ")"
+
+instance Eq (BaseField curve) => Eq (PointCompressed curve) where
```
TurtlePU, line 130: Can this be derived?

src/ZkFold/Base/Protocol/ARK/Plonk.hs
```diff
@@ -60,6 +62,17 @@ instance (KnownNat n, KnownNat l, Arithmetic (ScalarField c1), Arbitrary (Scalar
         let (omega, k1, k2) = getParams (value @n)
         Plonk omega k1 k2 (Vector vecPubInp) ac <$> arbitrary
 
+instance forall n l c1 c2 t . (KnownNat n, KnownNat l, Eq (ScalarField c1), Arithmetic (ScalarField c1), Arbitrary (ScalarField c1),
```
TurtlePU, line 65: `Eq a` is a superclass of `Arithmetic a`, so this looks like a redundant constraint
Also it's strange that `Witness` type family doesn't compute, what does it say if one removes a constraint on witness?

TurtlePU (dismissed):
LGTM except some minor comments

src/ZkFold/Base/Protocol/ARK/Plonk.hs
vlasin: It says "could not deduce `Witness (Plonk n l c1 c2 t) ~ (PlonkWitnessInput c0, PlonkProverSecret b0)`

TurtlePU (approved):

PR #235, opened 2024-08-30
Upgrade Plonk to PlonkUp

vlasin:
Checklist:

- [x] One type - one module
- [x] Rename types to `Plonkup*`
- [x] Refactor tests
- [x] Add more tests for `plonkupProve`
- [X] Single lookup table case
- [x] Add a test to check the correctness of lookup queries
- [x] Clean up code
- [x] Plonk as a special case


vlasin:
API Changes:
- `prove` will now return non-negated witness components
- `verify` now expects non-negated public inputs

src/ZkFold/Base/Protocol/Plonkup/Prover/Secret.hs
```diff
@@ -0,0 +1,61 @@
...
+
+module ZkFold.Base.Protocol.Plonkup.Prover.Secret where
+
+import           GHC.Generics                            (Generic)
+import           Prelude                                 hiding (Num (..), drop, length, sum, take, (!!), (/), (^))
+import           Test.QuickCheck                         (Arbitrary (..))
+
+import           ZkFold.Base.Algebra.EllipticCurve.Class (EllipticCurve (..))
+
+data PlonkupProverSecret c = PlonkupProverSecret {
```
TurtlePU, line 11: Maybe a single `b :: Vector 19 (ScalarField c)` can be used instead?

src/ZkFold/Base/Protocol/Plonkup/Internal.hs
```diff
@@ -3,227 +3,47 @@
...
+    NOTE: we need to parametrize the type of transcripts because we use BuiltinByteString on-chain and ByteString off-chain.
+    Additionally, we don't want this library to depend on Cardano libraries.
+-}
+
+data Plonkup (i :: Natural) (n :: Natural) (l :: Natural) curve1 curve2 transcript = Plonkup {
+        omega :: ScalarField curve1,
+        k1    :: ScalarField curve1,
+        k2    :: ScalarField curve1,
+        xPub  :: Vector l (Var (Vector i)),
+        ac    :: ArithmeticCircuit (ScalarField curve1) (Vector i) Par1,
```
TurtlePU, line 29: Consider changing the output type to `Vector l` to incorporate public variables as the `acOutput` part of a circuit instead of a separate field in `Plonkup` datatype

TurtlePU (approved):
🔥 🔥 🔥

Issue #222, opened 2024-08-18
Improve `WitnessField` definition


TurtlePU:
Turns out that the [current definition](https://hackage.haskell.org/package/zkfold-base-0.1.0.0/candidate/docs/ZkFold-Symbolic-MonadCircuit.html#t:WitnessField) of a `WitnessField` is not general enough for our usecases. Most importantly, it does not support conversion of witnesses to some "unbounded natural" type, which is necessary for efficient implementation of foreign-field arithmetic; in addition, witness construction in `splitExpansion` can be simplified. Proposed new definition:
```
type WitnessField n a = (FiniteField a, FromConstant n a, ToConstant a n, EuclideanDomain n)
```
The `binaryExpansion` can be recovered by converting `a` to `n`, iteratively calling `divMod` and converting results back to `a`.
Also this way definition of `Arithmetic` becomes more natural: with `n ~ Natural`, `Eq` and `Ord` are both inherited from `Natural`.

PR #238, opened 2024-09-02
Made ToConstant singleparam

TurtlePU:
Made `ToConstant` into a single-parameter class with an associated `Const` type. In correspondence with #154 and useful for simpler implementation of #222.


echatav:
I'm wary of this change.

* Haskell's `toInteger :: Integral a => a -> Integer` and `toRational :: Real a => a -> Rational` that `toConstant` abstracts, the `b` type is not _determined_ by `a`.
* In an algebraic sense `toConstant` is "there is some natural or standard injection of `a` into `b`".* Fixing the types removes composability of natural injections, which have good properties.
* Adds a type family

(*which is why I renamed it `to :: Into a b => a -> b` in hierarchy proposal)

TurtlePU:
Consider the following definition
```haskell
to' :: forall b a. (ToConstant a, FromConstant (Const a) b) => a -> b
to' = fromConstant . toConstant
```
(I was naming it `cast`, but let's call it `to'` as it is actually the same function that you consider)
* Both `to' @Integer` and `to' @Rational` can be inferred from single `ToConstant a` with `Const a ~ Integer`
* are associated types problematic? I would like to know more as I was using them here and there, maybe we would need to change these places...
* `to' @a @a = id` can be a property of `ToConstant` class

echatav:
But then you're mixing `FromConstant` in, which is a different sort of natural morphism together with the assumed property that `to . from = id`.

In #154, the concept is that `from :: From x a => x -> a` is a natural projection onto a subset, like projection of free `Semiring Natural`, `Ring Integer` and `Field Rational` onto the sub-{semiring,ring,field} generated by 0 & 1 and that `to :: Into y a => y -> a` is as I said a natural injection. The properties are similar but because one is an algebraic projection and the other is an inclusion, they're kept separate and separately compositional.

```
-- prop> from . from = from
-- prop> from @a @a = id @a
-- prop> to . to = to
-- prop> to @a @a = id @a
-- prop > to . from = id
```

echatav:
This conversation has been resolved ^

TurtlePU:
@vlasin what do you think?

echatav (approved):
👍

vlasin (approved):

Issue #223, opened 2024-08-18
Change variable index generation algorithm for arithmetic circuits


TurtlePU:
As of now, a variable index generation algorithm is an elaborate mix of evaluation of a constraint polynomial and taking dependencies of a new variable into account. This can be simplified by only evaluating a witness argument with a tricky `WitnessField` instance: elements of this "field" are Merkle hashes, and "field operations" are actually hashing algorithms run on their arguments (and on the "operation identifier" we can specify arbitrarily). New variable is this very merkle hash (maybe paired with a "distance from input variables" component to reduce probability of collisions).

In addition, this improves the story with `newRanged` where I still do not like that we generate this "pseudo-constraint-polynomial" just to stick into the old variable generation algorithm.

Even better, this way we can completely disentangle variable generation from constraint generation and simplify the `MonadCircuit` API:
```haskell
class MonadCircuit v a m | m -> v, m -> a where
  unconstrained :: Witness a v -> m v
  constraint :: ClosedPoly a v -> m ()
  rangeConstraint :: v -> a -> m ()

  new :: ClosedPoly a v -> m v
  new p = do
    v <- unconstrained p
    constraint (\x -> p x - x v)
    return v
```

TurtlePU:
For hashing, we can use https://hackage.haskell.org/package/cryptohash-sha256.
Pros:
* said to be fast
* does not introduce any new dependencies (we already depend both on `base` and `bytestring`)
* we can reuse the `Binary` machinery to constrain `Rep`s used as variables (but due to the nature of `Rep`s we have, maybe we need to do something a little smarter)
Cons:
* duplicates the code: we already have our own `SHA2` (only for symbolic types, though)

PR #244, opened 2024-09-04
Circuit cleanup

TurtlePU:
Cleanup of a circuit-related code before resolving #223
* Consistent set of fields used in `ArithmeticCircuit.Instance` module: `acRange`, `acSystem`, `acOutput`
* Removed practically unused `acVarOrder` field
* Removed `(0, one)` entry from starting witness, now it's empty
* Updated `PlonkConstraint` variable type to `Maybe (Var i)` to properly model the absence of variable


echatav (approved):

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
```diff
@@ -40,48 +42,47 @@ instance (Arbitrary a, Finite a, ToConstant a Natural, KnownNat i) => Arbitrary
         qr <- arbitrary
         qo <- arbitrary
         qc <- arbitrary
-        x1 <- NewVar . toConstant @a <$> arbitrary
-        x2 <- NewVar . toConstant @a <$> arbitrary
-        x3 <- NewVar . toConstant @a <$> arbitrary
+        x1 <- Just . NewVar . toConstant @a <$> arbitrary
```
vlasin, line 45: It makes arbitrary less general than before. Can we include the possibility of constants as well?

TurtlePU: Are you talking about the case of `NewVar 0`? We can do it, but what should we set the probability of this to? Something like 1/20?

src/ZkFold/Base/Protocol/Plonkup/Relation.hs
```diff
@@ -61,9 +63,13 @@ toPlonkRelation xPub ac0 =
         -- TODO: Permutation code is not particularly safe. We rely on the list being of length 3*n.
         sigma = fromCycles @(3*n) $ mkIndexPartition $ fromList $ a ++ b ++ c
 
-        w1 i   = toPolyVec $ fromList $ map (indexW ac i) a
-        w2 i   = toPolyVec $ fromList $ map (indexW ac i) b
-        w3 i   = toPolyVec $ fromList $ map (indexW ac i) c
+        indexW _ Nothing           = one
```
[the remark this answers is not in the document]
TurtlePU: Because this is the replacement for previous `NewVar 0` which was used as "constant 1 input"

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
vlasin: Hmm, yes. I don't think it is possible to generate a constraint with less than 3 variables, like `x1 + x2 == 0`. I think equal probabilities for generating a constraint with `k` variables (k=1,2,3) would be great.

vlasin (approved):

PR #245, opened 2024-09-05
Update `WitnessField`

TurtlePU:
This PR updates `WitnessField` typealias by replacing `BinaryExpansion` constraint with a conversion to the "unbounded natural numbers" type which is a `EuclideanDomain`. Also some minor improvements along the way.

Branches out from #244, so maybe that one has to be merged first.

### DIFF
--- a/src/ZkFold/Base/Algebra/Basic/Class.hs
+++ b/src/ZkFold/Base/Algebra/Basic/Class.hs
@@ -269,16 +269,20 @@ intScale n a | n < 0     = naturalFromInteger (-n) `scale` negate a
 -}
 class (AdditiveMonoid a, MultiplicativeMonoid a, FromConstant Natural a) => Semiring a
 
-{- | A Euclidean domain @R@ is an integral domain which can be endowed
-with at least one function @f : R\{0} -> R+@ s.t.
-If @a@ and @b@ are in @R@ and @b@ is nonzero, then there exist @q@ and @r@ in @R@ such that
-@a = bq + r@ and either @r = 0@ or @f(r) < f(b)@.
+{- | A semi-Euclidean-domain @a@ is a semiring without zero divisors which can
+be endowed with at least one function @f : a\{0} -> R+@ s.t. if @x@ and @y@ are
+in @a@ and @y@ is nonzero, then there exist @q@ and @r@ in @a@ such that
+@x = qy + r@ and either @r = 0@ or @f(r) < f(y)@.
 
-@q@ and @r@ are called respectively a quotient and a remainder of the division (or Euclidean division) of @a@ by @b@.
+@q@ and @r@ are called respectively a quotient and a remainder of the division
+(or Euclidean division) of @x@ by @y@.
 
-The function @divMod@ associated with this class produces @q@ and @r@ given @a@ and @b@.
+The function @divMod@ associated with this class produces @q@ and @r@
+given @a@ and @b@.
+
+This is a generalization of a notion of Euclidean domains to semirings.
 -}
-class Semiring a => EuclideanDomain a where
+class Semiring a => SemiEuclidean a where
     {-# MINIMAL divMod | (div, mod) #-}
 
     divMod :: a -> a -> (a, a)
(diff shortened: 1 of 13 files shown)


TurtlePU:
Tests fail for `UInt`'s `Ord` instance and `divMod`, investigating
Most probably something to do with variable generation again

TurtlePU:
@vlasin, all tests pass except "Conversion to Plonk constraints and back" test because in case when some of variables are `Nothing`, constraints are not equal (and don't have to be, according to the definition of a constraint)
So this is not in scope of this PR but has to be fixed in a separate PR

TurtlePU:
@echatav, can you take a look at this PR? It is ready.

src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -279,9 +279,10 @@ If @a@ and @b@ are in @R@ and @b@ is nonzero, then there exist @q@ and @r@ in @R
 The function @divMod@ associated with this class produces @q@ and @r@ given @a@ and @b@.
 -}
 class Semiring a => EuclideanDomain a where
```
echatav, line 281: Let's rename this to `SemiEuclidean` in this PR? `Euclidean(Domain)` should be `Ring` + `SemiEuclidean` + `eea`

echatav (changes requested):
Looks good to me, but let's take the time now to rename one class?

echatav (approved):
:shipit:

src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -269,19 +269,24 @@ intScale n a | n < 0     = naturalFromInteger (-n) `scale` negate a
 -}
 class (AdditiveMonoid a, MultiplicativeMonoid a, FromConstant Natural a) => Semiring a
 
-{- | A Euclidean domain @R@ is an integral domain which can be endowed
-with at least one function @f : R\{0} -> R+@ s.t.
-If @a@ and @b@ are in @R@ and @b@ is nonzero, then there exist @q@ and @r@ in @R@ such that
-@a = bq + r@ and either @r = 0@ or @f(r) < f(b)@.
+{- | A semi-Euclidean-domain @a@ is a semiring without zero divisors which can
+be endowed with at least one function @f : a\{0} -> R+@ s.t. if @x@ and @y@ are
```
echatav, line 273: Screenshot 2024-09-06 at 2 47 49 PM

vlasin:
> @vlasin, all tests pass except "Conversion to Plonk constraints and back" test because in case when some of variables are `Nothing`, constraints are not equal (and don't have to be, according to the definition of a constraint) So this is not in scope of this PR but has to be fixed in a separate PR

Could you elaborate? Why don't they have to be equal?

TurtlePU:
> Could you elaborate? Why don't they have to be equal?

Because if one of the inputs is constant, evaluating Plonk constraint into polynomial loses information which one of them. For example,

1. The constraint `C` is `1 x c(1) + 0 x + 0 c(1) + (-1) z + 0 = 0` where `c(1)` is the placeholder constant input
2. Evaluated as a polynomial, this is `x - z = 0`
3. As a Plonk constraint, this can be represented:
  * as `C`
  * as `0 x c(1) + 1 x + 0 c(1) + (-1) z + 0 = 0`
  * as `0 c(1) x + 0 c(1) + 1 x + (-1) z + 0 = 0`
  * as `0 c(1) z + 0 c(1) + (-1) z + 1 x + 0 = 0`
  * as `0 z c(1) + (-1) z + 0 c(1) + 1 x + 0 = 0`
  * also in a myriad more ways as `a x c(1) + (1 - a) x + 0 c(1) + (-1) z + 0 = 0`
 
Before #244, the possibility of introducing `c(1)` in arbitrary Plonk constraint was negligible, so this was not caught before. Also, analogously, the same issue arises with linear constraints of the form `x + y - z = 0` because there are 6 options how this constraint can be represented, but this one again has a negligible probability in our tests because for this, we need to have the coefficient near `xy` to equal `0`.

vlasin:
Ok, I agree to fix it in another PR.

Issue #60, opened 2024-03-28
Add input validation for arithmetizable functions


vlasin:
Currently, the inputs of arithmetizable computations are not validated. The simplest example is the `bool` function. It can be arithmetized in three Plonk constraints: `bool x y (Bool b) == b * (y - x) + x`. To validate `b`, we would need an extra constraint `b * (one - b) == zero`.

If we validate the inputs of every intermediate operation, we will be adding a lot of duplicate constraints to the circuit. So, I suggest adding input validation constraints in the `Arithmetizable` instance for functions, because it only runs for the top-level functions, and the arguments of these functions are the true inputs to the resulting circuits.

TurtlePU:
With
* definition of `Symbolic` from #195 
* definition of `SymbolicData` from [here](https://github.com/zkFold/zkfold-base/issues/102#issuecomment-2250016689) 
* and with a version of @echatav's structured inputs from #177,

we can do something along these lines:
```haskell
class (SymbolicData c d, Input c d ~ (), Representable (Layout c d)) => SymbolicInput c d where
  inputConstraint :: MonadCircuit i (BaseField c) m => Layout c d i -> m ()
  inputConstraint = return () -- default impl imposes no constraints

class RepCat c where
  rid :: Representable i => c i i
  -- can also have rcomp :: c j k -> c i j -> c i k, but this is not needed here

-- instance RepCat (AC a) where output rid = tabulate InVar

input :: (RepCat c, Symbolic (c i), SymbolicInput (c i) d, Layout (c i) d ~ i) => d
input = restore $ const $ fromCircuitF rid (\is -> inputConstraint is >> return is)
```

TurtlePU:
A case study for `inputConstraint` proposed above: `Maybe` symbolic type.

The desired input constraints look like: "either discriminant is false (and value is `Nothing`) or discriminant is true and the value inside satisfies the input constraints". This cannot be expressed using the definition of `inputConstraint` above in a compositional way. Maybe instead it should have the following type:
```haskell
inputConstraint :: Layout c d i -> System (BaseField c) i
```

Where `System a i` is a CNF of explicit (polynomial? range?) constraints which is a `BoolType` and can be composed into an arbitrarily complex formula which is then desugared into the system of constraints during compilation.

However, this way we cannot create new intermediate variables while constructing the system. Is it desired? Is there a type which needs additional variables to encode its variants?

vlasin:
[...]

PR #248, opened 2024-09-11
Constant Variables

### LINKED ISSUE #65: Optimize `bool` and constant embedding

echatav:
Resolves #65 

Modifies circuit `Var` type so that now variables are classified like so:
- system variables
  - input variables
  - new assigned variables
- constant variables

Circuit constraints only use system variables so that constant variables must be evaluated, optimizing constants. Circuit outputs may use either system or constant variables, so embedding of constants does not require new variable assignment.

Added benchmarks for constants demonstrates optimization:
```
Constant.5
Number of constraints: 0
Number of variables: 0
Number of range lookups: 0

Eq.Constant.5
Number of constraints: 4
Number of variables: 4
Number of range lookups: 0

Eq
Number of constraints: 4
Number of variables: 4
Number of range lookups: 0
```
The constant circuit "5" has zero constraints.
The circuit "equals 5" has the same number of constraints as "equals".
The number of variables means _system_ variables, since constants only appear in circuit outputs they "don't count".
The number of variables and constraints were reduced for some of the benchmark examples.


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -66,27 +70,47 @@ data ArithmeticCircuit a i o = ArithmeticCircuit
...
+acInput0 :: Representable i => i (SysVar i)
+acInput0 = fmapRep InVar (tabulate id)
 
-data Var i
+acInput :: Representable i => i (Var a i)
+acInput = fmapRep (SysVar . InVar) (tabulate id)
+
+------------------------------ Symbolic compiler context ----------------------------
+
+data SysVar i
```
TurtlePU, line 87: Maybe `Var` definitions should not be in the section "Symbolic compiler context", but next to the `data ArithmeticCircuit` declaration

TurtlePU (dismissed):
LGTM but we need to decide whether to merge this or #235 first as they conflict with one another

TurtlePU:
(Pinging @vlasin too)

vlasin:
I propose to merge #235 first, as I have to move on to other tasks now.

Questions:

1. Does it solve #65?
2. What else is needed?
3. Does it reduce the constraint numbers in our benchmarks?

echatav:
1. From my PoV, this resolves #65 
2. For #65 I don't think anything else will be needed, pending (3.)
3. I need to investigate this.

If another conflicting PR is being merged first then we can take some time on this one.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -98,12 +120,20 @@ witnessGenerator circuit inputs =
     in
         result
 
+indexW :: Representable i => ArithmeticCircuit a i o -> i a -> Var a i -> a
```
TurtlePU, line 123: Note for future: maybe we can create a helper datatype like this:
```haskell
data Witness a i = Witness { wInput :: i a, wNew :: Map Natural a }
instance Representable i => Representable (Witness a i) where
  type Rep (Witness a i) = Var a i
```
This would reduce clutter around witnesses a little.

echatav: I don't think this is `Representable` because it's not fixed-size/shape, even though `i` is. In particular `Map Natural` is not `Representable`.

TurtlePU: Oh yeah, sorry
I was thinking abt creating the empty map in `tabulate`, but I guess it doesn't satisfy the laws of `Representable`

TurtlePU (approved):

vlasin:
Having errors in `DivMod` and `greater than` tests for `UInt 32`.

TurtlePU:
Also a bunch of benchmark values changed, we need to run `compiler-benchmark` benchmark with `--accept` option to update the values in `stats/` folder

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
```diff
@@ -34,41 +33,39 @@ data PlonkConstraint i a = PlonkConstraint
...
         qr <- arbitrary
         qo <- arbitrary
         qc <- arbitrary
-        k <- elements [1, 2, 3]
-        xs0 <- sort <$> replicateA k (Just . NewVar . toByteString @a <$> arbitrary)
-        let (x, y, z) = case replicate (3 -! k) Nothing ++ xs0 of
-              [x', y', z'] -> (x', y', z')
-              _            -> error "impossible"
-        return $ PlonkConstraint qm ql qr qo qc x y z
+        let arbitraryNewVar = SysVar . NewVar . toByteString @a <$> arbitrary
```
echatav, line 49: Is it okay if arbitrary x1,x2,x3 are only new assigned variables, not input or constant variables?

echatav:
> Having errors in `DivMod` and `greater than` tests for `UInt 32`.

Verified these tests now pass with new variable generation algorithm.

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
echatav: In particular no `ConstVar one`s are generated...

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -128,19 +156,29 @@ instance
...
 instance
   ( Arithmetic a, Binary a, Representable i, Binary (Rep i), Ord (Rep i)
-  , o ~ U1) => MonadCircuit (Var i) a (State (ArithmeticCircuit a i o)) where
-
-    unconstrained witness = do
-      let v = toVar @a witness
+  , o ~ U1) => MonadCircuit (Var a i) a (State (ArithmeticCircuit a i o)) where
+
+    unconstrained witness' = do
+      let witness :: Witness (SysVar i) a
```
TurtlePU, line 162: `witness` definition can be inlined to its only usage to simplify implementation

echatav:
BatchTransfer benchmark example is too slow to run on my laptop.

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
TurtlePU: @vlasin

TurtlePU:
> BatchTransfer benchmark example is too slow to run on my laptop.

On mine too! Working on it

TurtlePU (dismissed):

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -168,10 +203,11 @@ instance
...
 toVar ::
   forall a i. (Finite a, Binary a, Binary (Rep i)) =>
-  Witness (Var i) a -> ByteString
+  Witness (Var a i) a -> ByteString
 toVar witness = runHash @(Just (Order a)) $ witness $ \case
-  InVar inV -> merkleHash inV
-  NewVar newV -> M newV
+  SysVar (InVar inV) -> merkleHash inV
+  SysVar (NewVar newV) -> M newV
+  ConstVar cV -> merkleHash cV
```
echatav, line 210: ^ Is there any chance that these witness variables could clash, like if `toByteString inV == toByteString cV`?

TurtlePU:
On `bool` and constant embedding:
* we can optimize constants further -- we can inspect the witness given to us in `unconstrained` and see if it is constant, then we can return `ConstVar` instead of `NewVar`
* we will need some kind of boolean optimization when we start working on #60
But for now I guess we can close the corresponding issue

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: Oh, actually there is a good chance

TurtlePU: We can add a "from constant" case to `Prec` and use it in `fromConstant` instance which we can call here

echatav:
> On `bool` and constant embedding:
> 
> * we can optimize constants further -- we can inspect the witness given to us in `unconstrained` and see if it is constant, then we can return `ConstVar` instead of `NewVar`
> * we will need some kind of boolean optimization when we start working on [Add input validation for arithmetizable functions #60](https://github.com/zkFold/zkfold-base/issues/60)
>   But for now I guess we can close the corresponding issue

Can you make new issues for these?

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: Well, not a really probable case (we'd need input vars to be Zp which need same amount of bits to encode as `a`), but still would be great to distinguish between variables and constants

echatav: I added a `Const` label to the `Prec` enum but I wonder if it should also have an `Input` label for input variables too 🤔

TurtlePU (approved):

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: I don't think so, the code looks clean to me and I don't think we'd encounter any collisions soon (not before sha256 collision is discovered ;))

src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
vlasin: I think it is fine. In general, I think we should get rid of this instance and just use Arbitrary for ArithmeticCircuit.

vlasin:
> * we will need some kind of boolean optimization when we start working on #60

Could you explain?

TurtlePU:
> Could you explain?

@vlasin, if we go with [`isValid` approach](https://github.com/zkFold/zkfold-base/issues/60#issuecomment-2267667699), constraining the input would go like this:

* call `isValid` on input
* call `forceOne` on the result of `isValid`

This way, we get a constraint `k - 1 == 0` in a circuit where `k` is a variable corresponding to the result of `isValid` call. This can be optimized "backwards":

1. If `isValid` was `not x`, there is a constraint `one - l - k == 0` where `l` corresponds to `x`. This can be simplified to `l == 0`, removing `k` from the circuit.
2. If `isValid` was `x && y`, there is a constraint `l * m - k == 0` where `l` corresponds to `x` and `m` corresponds to `y`. This can be simplified to `l * m - 1 == 0`; **knowing that `l` and `m` are boolean**, this simplifies further to `l - 1 == 0` and `m - 1 == 0`.
3. Further, if `isValid` was `x <= y`, it is a range constraint which could also be optimized somehow.

vlasin:
@TurtlePU, I see what you mean and agree that we could optimize input validation a bit with this. However, if I understand correctly, it will only reduce the number by a few constraints. So maybe it is not a priority.

TurtlePU:
Agreed, definitely not a priority

PR #260, opened 2024-09-16
Drop bs-aeson-orphans; use our combinators

TurtlePU:
This PR
* drops dependency on `bytestring-aeson-orphans` and adds our own `FromJSON`/`ToJSON` instances for `ByteString`
* reuses existing combinators from `ZkFold.Base.Data.ByteString` instead of `toStrict . encode` etc. in Circuit-related code


TurtlePU:
Running tests locally now to make sure everything is ok

vlasin (approved):

PR #261, opened 2024-09-17
Novel variable generation algorithm

TurtlePU:
Migrates `ArithmeticCircuit` to the new variable generation algorithm. Now, instead of `ToConstant (Rep i)`, we need `Binary a` and `Binary (Rep i)`.

Also compilation got a whole lot faster (because of efficient hashing used for var gen) and compiled circuits have less constraints (because variables are now in some sense content-addressable and, for example, 0s now all use the same variable).


src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -22,35 +23,32 @@ module ZkFold.Symbolic.Compiler.ArithmeticCircuit.Internal (
...
+import           ZkFold.Base.Algebra.Polynomials.Multivariate          (Poly, evalMonomial, evalPolynomial, var)
 import           ZkFold.Base.Control.HApplicative
-import           ZkFold.Base.Data.ByteString                  (fromByteString, toByteString)
 import           ZkFold.Base.Data.HFunctor
 import           ZkFold.Base.Data.Package
 import           ZkFold.Symbolic.Class
+import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.MerkleHash
 import           ZkFold.Symbolic.MonadCircuit
 
 -- | The type that represents a constraint in the arithmetic circuit.
```
vlasin, line 54: Do we still need `acRNG` field? Could you please describe somewhere the new algorithm for variable generation?

TurtlePU: No, we don't need `acRNG`, was going to remove it along with introducing the derived instance of `Monoid` for `ArithmeticCircuit` instead of handwritten in the next PR

TurtlePU: About algorithm description -- ok, will add it as a doc comment of `toVar` function.

src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -73,9 +71,8 @@ data ArithmeticCircuit a i o = ArithmeticCircuit
 deriving instance (NFData a, NFData (o (Var i)), NFData (Rep i))
     => NFData (ArithmeticCircuit a i o)
 
--- | A finite field of a large order.
--- It is used in the compiler for generating new variable indices.
-type VarField = Zp BLS12_381_Scalar
+-- | Variables are SHA256 digests (32 bytes)
+type VarField = Zp (2 ^ (32 * 8))
```
echatav, line 75: Is `VarField` still in use? It looks like some code referring to it was removed.

vlasin (approved):

PR #279, opened 2024-09-25
Use Data.Vector.Vector in Vector n

### LINKED ISSUE #262: Optimize Vector type by using Data.Vector instead of linked lists.
Optimize Vector type by using Data.Vector instead of linked lists.

echatav:
Resolves #262


src/ZkFold/Base/Data/Vector.hs
```diff
@@ -5,110 +5,104 @@
...
-import           ZkFold.Prelude                   (length, replicate)
+import           ZkFold.Prelude                   (length)
 
-newtype Vector (size :: Natural) a = Vector [a]
+newtype Vector (size :: Natural) a = Vector {toV :: V.Vector a}
     deriving (Show, Eq, Functor, Foldable, Traversable, Generic, NFData, Ord)
 
+-- helper
+knownNat :: forall size n . (KnownNat size, Integral n) => n
+knownNat = knownNat @size
```
TurtlePU, line 34: I feel like this is going to loop

src/ZkFold/Base/Data/Vector.hs
```diff
@@ -5,110 +5,104 @@
...
-drop (Vector lst) = Vector (ZP.drop (value @n) lst)
+drop (Vector lst) = Vector (V.drop (knownNat @n) lst)
 
 splitAt :: forall n m a. KnownNat n => Vector (n + m) a -> (Vector n a, Vector m a)
-splitAt (Vector lst) = (Vector (ZP.take (value @n) lst), Vector (ZP.drop (value @n) lst))
+splitAt (Vector lst) = (Vector (V.take (knownNat @n) lst), Vector (V.drop (knownNat @n) lst))
 
 -- | The sole purpose of this function is to get rid of annoying constraints in ZkFols.Symbolic.Compiler.Arithmetizable
 --
 splitAt3 :: forall n m k a. (KnownNat n, KnownNat m) => Vector (n + m + k) a -> (Vector n a, Vector m a, Vector k a)
```
TurtlePU, line 101: If I remember correctly, this function is unused and we can remove it

src/ZkFold/Base/Data/Vector.hs
```diff
@@ -119,71 +113,60 @@ rotate (Vector lst) n = Vector (r <> l)
...
-unsafeConcat = Vector . concatMap fromVector
+unsafeConcat = concat . unsafeToVector @m
 
 chunks :: forall m n a . KnownNat n => Vector (m * n) a -> Vector m (Vector n a)
-chunks (Vector lists) = Vector (Vector <$> chunksOf (fromIntegral $ value @n) lists)
+chunks (Vector vectors) = unsafeToVector (Vector <$> V.chunksOf (fromIntegral $ value @n) vectors)
 
 instance Binary a => Binary (Vector n a) where
     put = put . fromVector
     get = Vector <$> get
```
TurtlePU, line 149: There is no guarantee that the parsed ByteString would contain exactly `n` elements, something like `replicate n` is needed here

src/ZkFold/Symbolic/Cardano/Types/Value.hs
```diff
@@ -36,17 +35,3 @@ deriving instance
 
 instance Symbolic context => Scale Natural (Value n context) where
     n `scale` Value v = Value $ fmap (\((pid, aname), q) -> ((pid, aname), n `scale` q)) v
-
-instance (Haskell.Ord (PolicyId context), Haskell.Ord (AssetName context), Symbolic context) => Semigroup (Value n context) where
```
TurtlePU, line 40: Why are these instances removed?

src/ZkFold/Base/Data/Vector.hs
echatav: Yup, I think it suffered a find & replace mishap. Fixed.

src/ZkFold/Base/Data/Vector.hs
echatav: removed

src/ZkFold/Symbolic/Cardano/Types/Value.hs
echatav: I can put them back but to be honest they didn't make much sense to me, so I ended up removing them when making modifications. The `mempty`/`zero` value has 0 length even though `Value` is a newtype over `Vector n`. That can't be right, right?

src/ZkFold/Base/Data/Vector.hs
echatav: fixed it to use V.replicateM

src/ZkFold/Symbolic/Cardano/Types/Value.hs
TurtlePU: Yeah, now that you mention it, looks strange to me.
Pinging @vlasin to take a look at it too. Maybe Symbolic `List` was expected to be used here instead of `Vector`?

TurtlePU (dismissed):

src/ZkFold/Symbolic/Cardano/Types/Value.hs
vlasin: Yes, exactly. `Value` is a "basket of assets". We should be able to merge two baskets into one. Using `Vector n` for this is less than ideal, but we don't have the Symbolic List just yet.

vlasin:
`Data.Vector.Vector` is used directly in univariate polynomials. If we could use it through `ZkFold.Base.Data.Vector`, it might clean up the code over there. Maybe in a future PR?

src/ZkFold/Symbolic/Cardano/Types/Value.hs
echatav: But we could use functions with more sensible signatures. Shouldn’t combining add the lengths. And shouldn’t empty value be length zero?

TurtlePU: The most sensible signatures would be when we use Symbolic lists here instead of Vectors, for now it is better to keep instances in place for future reference when we do the migration

TurtlePU: (Also combining does not necessarily add the lengths, the resulting length might be smaller because of equal keys)

vlasin: Yes, for the time being, we could think of `n` as the maximal possible size of the basket. So, an "overflow" will result in a system with an empty solution set.

echatav: Added instances back.

src/ZkFold/Base/Data/Vector.hs
```diff
@@ -5,119 +5,109 @@
...
+vtoVector as
+  | V.length as == knownNat @size = Just $ Vector as
+  | otherwise                     = Nothing
+
 instance IsList (Vector n a) where
     type Item (Vector n a) = a
     toList = fromVector
     fromList = unsafeToVector
 
-parFmap :: (a -> b) -> Vector size a -> Vector size b
```
TurtlePU, line 50: Btw why it doesn't error out? I thought we were using it somewhere

Also maybe we can add some analog of it back?

echatav: It’s not used anywhere

TurtlePU (approved):

Issue #239, opened 2024-09-03
ZK KYC: data types and a KYC check example in Symbolic


vlasin:
Add types related to the ZK KYC application to zkFold Symbolic. See this [document](https://github.com/DefyFinancial/defy-zkfold-cross-border-payment/tree/main/src/11-record-kyc-data-ready-for-zk-operations) for details.

Checklist for this issue:

- [ ] All data types and type aliases/newtypes are implemented in Symbolic
- [ ] All data types have JSON instances
- [ ] Create a function in Symbolic that represents a KYC check (a complex boolean expression that uses multiple data fields of different types).

vlasin:
@vks4git I added an example of using Symbolic in this kind of application as a part of this issue. Hope you don't mind :)

TurtlePU:
> All data types have JSON instances

How are `ToJSON` instances of Symbolic datatypes going to be used, btw?

vlasin:
> > All data types have JSON instances
> 
> How are `ToJSON` instances of Symbolic datatypes going to be used, btw?

Do you mean for circuits? We only need them for the interpreter version.

TurtlePU:
> We only need them for the interpreter version.

Oh, okay, nice!

vlasin:
However, we might want a circuit serialization feature at some point. Also, conversions to/from other popular formats.

TurtlePU:
Then I think we can model this more accurately with the help of #308, like this:
```haskell
fromValueJSON :: (SymbolicValue d, FromJSON (Value d)) => ... 
toValueJSON :: (SymbolicValue d, ToJSON (Value d), Context d ~ c, BaseField c ~ WitnessField c) => ...
```
And derive `FromJSON`/`ToJSON` for Symbolic data which would give us circuit serialization behavior

PR #282, opened 2024-09-30
Add a KYC datatype

charrsky:
#239


echatav (changes requested):

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -355,3 +356,11 @@ instance (Symbolic c, NumberOfBits (BaseField c) ~ n) => Iso (FieldElement c) (B
...
 instance (Symbolic c, NumberOfBits (BaseField c) ~ n) => Iso (ByteString n c) (FieldElement c) where
   from (ByteString a) = fromBinary a
+
+instance (FromConstant Natural (ByteString 8 c), Concat (ByteString 8 c) (ByteString n c))
+    => FromJSON (ByteString n c) where
+    parseJSON = Haskell.fmap fromString . parseJSON @String
+
+instance (ToConstant (ByteString n (Interpreter (Zp p))))
+    => ToJSON (ByteString n (Interpreter (Zp p))) where
+    toJSON = toJSON . toConstant
```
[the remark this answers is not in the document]
charrsky: well, I'm not sure how to convert our symbolic bytestring into just string, so I left it as is for now

src/ZkFold/Symbolic/Data/KYC.hs
```diff
@@ -0,0 +1,25 @@
...
+
+data KYCData context = KYCData
+    { kycType  :: KYCByteString context
+    , kycValue :: KYCByteString context
+    , kycHash  :: KYCByteString context
+    , kycID    :: UInt 64 Auto context
+    } deriving Generic
+
+isCitizen :: (Symbolic c) => KYCByteString c -> Vector n (KYCByteString c) -> Bool c
+isCitizen citizenship allowedCountries = foldl' (||) false (parFmap (== citizenship) allowedCountries)
```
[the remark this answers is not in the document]
charrsky: I tried but then got some type error. Just replaced my foldl'-definition with `ZkFold.Symbolic.Data.Bool`'s any, thanks!

src/ZkFold/Symbolic/Data/ByteString.hs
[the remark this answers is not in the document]
echatav: We must maintain the invariant property `Aeson.decode . Aeson.encode = Just`

echatav (changes requested):
Verify invariant JSON property.

TurtlePU:
@charrsky, for resolving the issue @echatav mantioned, you can convert from/to JSON via Haskell's `ByteString` type. It is `base64`-encoded and should work properly.

src/ZkFold/Symbolic/Data/KYC.hs
```diff
@@ -0,0 +1,24 @@
...
+
+data KYCData context = KYCData
+    { kycType  :: KYCByteString context
+    , kycValue :: KYCByteString context
+    , kycHash  :: KYCByteString context
+    , kycID    :: UInt 64 Auto context
+    } deriving Generic
+
+isCitizen :: (Symbolic c) => KYCByteString c -> Vector n (KYCByteString c) -> Bool c
+isCitizen citizenship allowedCountries = any (== citizenship)
```
TurtlePU, line 24: How does this pass the typechecker? `any` is not applied to any `Foldable` as far as I can see
I guess `allowedCountries` argument is to be removed

src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -355,3 +357,11 @@ instance (Symbolic c, NumberOfBits (BaseField c) ~ n) => Iso (FieldElement c) (B
 
 instance (Symbolic c, NumberOfBits (BaseField c) ~ n) => Iso (ByteString n c) (FieldElement c) where
   from (ByteString a) = fromBinary a
+
+instance (FromConstant Natural (ByteString 8 c), Concat (ByteString 8 c) (ByteString n c))
```
TurtlePU, line 361: Much better, but now we still need some tests to make sure this is right
I'd propose to add a test to `Tests/ByteString.hs` module which checks that `parseJSON . toJSON = Just`

echatav: Please also add same JSON property test but for the `UInt`s too. These are good tests to have for these basic symbolic types.

echatav (approved):
JSON invariance property for a couple base symbolic types is now tested

symbolic-base/src/ZkFold/Symbolic/Data/KYC.hs
TurtlePU, line 1: @vlasin, should this be in `symbolic-base` or somewhere else?

TurtlePU (approved):

symbolic-base/src/ZkFold/Symbolic/Data/KYC.hs
vlasin: I propose to create a `symbolic-apps` package that will contain Symbolic code for this and our other side-projects.

PR #290, opened 2024-10-09
Vks4git/143 in circuit cryptography

vks4git:


symbolic-base/src/ZkFold/Symbolic/Data/Ed25519.hs
```diff
@@ -49,14 +44,14 @@ instance
...
-    pieces Inf         = hliftA2 V.append <$> pieces (zero :: UInt 256 'Auto c) <*> pieces (zero :: UInt 256 'Auto c)
+    pieces Inf         = hliftA2 V.append <$> pieces (zero :: FFA Ed25519_Base c) <*> pieces (zero :: FFA Ed25519_Base c)
     pieces (Point x y) = hliftA2 V.append <$> pieces x <*> pieces y
 
     restore f = Point x y
         where
             (x, y) = restore f
 
-instance (Symbolic c, Eq (Bool c) (BaseField (Ed25519 c))) => Eq (Bool c) (Point (Ed25519 c)) where
+instance (Symbolic c) => Eq (Bool c) (Point (Ed25519 c)) where
```
vlasin, line 54: Does this actually work for circuits? I think we might get some unexpected errors when we use it in the circuit context. For example, `Inf /= Point zero zero` even though it's the same point.

symbolic-base/src/ZkFold/Symbolic/Data/Ed25519.hs
```diff
@@ -49,14 +44,14 @@ instance
     -- It will need additional checks in pointDouble because of the denominator becoming zero, though.
     -- TODO: Think of a better solution
     --
-    pieces Inf         = hliftA2 V.append <$> pieces (zero :: UInt 256 'Auto c) <*> pieces (zero :: UInt 256 'Auto c)
+    pieces Inf         = hliftA2 V.append <$> pieces (zero :: FFA Ed25519_Base c) <*> pieces (zero :: FFA Ed25519_Base c)
```
vlasin, line 47: `retore (pieces Inf)` is not `Inf`

vks4git: There is no way to return `Inf` with the current interface. `bool` relies on `restore` which would rely on `bool` itself if we were to perform a check if the point is (0, 0). The only solution which will not use such things is completely removing Ed25519 from Symbolic and treating it separately each time we want to use it

vlasin: Ok, I guess we could leave it as is for now. In general, Symbolic cannot normally support sum types. I think at some point, we should change the `Point` type.

TurtlePU: I'd honestly define a separate "Symbolic Ed25519" datatype which is just a pair $(x, y)$ of affine coordinates
There's no need for it to be a `Point`, and we can reuse `addPoints` etc. from curves by just putting coordinates into `Point` datatype and matching the result back into Symbolic representation

TurtlePU: (And as for `Pairing` class, I'd generalize its definition to arbitrary groups as pairing is defined for arbitrary groups anyway. This would also simplify type signatures in `Plonk` quite a bit)

vlasin: Yes, on the second thought, I agree with both points.

symbolic-base/src/ZkFold/Symbolic/Data/FFA.hs
```diff
@@ -196,3 +208,5 @@ instance (Prime p, Symbolic c) => Field (FFA p c) where
 
 instance Finite (Zp p) => Finite (FFA p b) where
   type Order (FFA p b) = p
+
+deriving newtype instance Symbolic c => Eq (Bool c) (FFA p c)
```
TurtlePU, line 212: The problem with current implementation of FFA, among many (performance, inefficient implementation of inverses,...), is that it is imprecise: the value is encoded up to the `+ p k` for some `k`.
So this `Eq` instance is incorrect, to make it work, we need to basically rewrite `FFA` from scratch to the more efficient implementation which is now possible thanks to the new `MonadCircuit` interface: https://hackmd.io/@kullervo/cryptographyExplorations/%2F%40kullervo%2FbignumberCircuits

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
```diff
@@ -98,40 +99,71 @@ getBitsBE x =
...
             update :: (i, i) -> (i, i) -> m (i, i)
             update (z1, z2) (x, y) = do
-                z1' <- newAssigned $ \p -> p z1 + (one - p z1) * (one - p z2) * (one - p y) * p x
-                z2' <- newAssigned $ \p -> p z2 + (one - p z1) * (one - p z2) * (one - p x) * p y
+                -- @f1@ is one if and only if @x > y@ and zero otherwise.
+                -- @(y + 1) `div` (x + 1)@ is zero if and only if @y < x@ regardless of whether @x@ is zero.
+                -- @x@ and @y@ are expected to be of at most @r@ bits where @r << NumberOfBits a@, so @x + 1@ will not be zero either.
+                -- Because of our laws for @finv@, @q // q@ is 1 if @q@ is not zero, and zero otherwise.
+                -- This is exactly the opposite of what @f1@ should be.
+                f1 <- newRanged one $
```
TurtlePU, line 145: `f1` is underconstrained: nothing prevents the adversary from setting `f1` to arbitrary value in range `[0; 1]`

TurtlePU: (However witness construction is honestly genious)

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
```diff
@@ -98,40 +99,71 @@ getBitsBE x =
...
+                    \p -> let q = fromConstant $ (toConstant (p y + one) `div` toConstant (p x + one))
+                           in one - q // q
+
+                -- f2 is one if and only if y > x and zero otherwise
+                f2 <- newRanged one $
+                    \p -> let q = fromConstant $ (toConstant (p x + one) `div` toConstant (p y + one))
+                           in one - q // q
+
+                d1  <- newAssigned (\p -> p f1 * (p x - p y - one))
```
TurtlePU, line 154: And this creates a non-Plonk constraint (the previous solution was also not Plonk though)

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
vks4git: Wait, how do range constraints work, then? `f1` was created with `newRanged`, I thought it will create a record for `f1` in `acRange`, won't it?

TurtlePU: It will, but this will only constrain `f1` to be in range `[0; 1]`, this does not force it to be a certain value

vks4git: Oh, true, but does it matter? Take a look at `d1` and `d1'`. They have both assigned values based on `f1` and range constraints. If `f1` is set incorrectly, constraints for `d1` and `d1'` won't be satisfied. E.g. if `f1` is zero while `x > y`, `d1` will be set to `y - x` which is outside the range `[0; 2^r]`. Is it still considered unsafe or underconstrained?

TurtlePU: Oh, wow
Yeah, this works as long $2^{r+1}\le |a|$

TurtlePU: Also note that this is performed on bits, I think you meant `x` and `y` to be arbitrary field elements so we can avoid `binaryExpansion` in `Ord`

vks4git: We pick `r` in code, so as long as we are careful, it's going to work properly

vlasin:
> @vlasin, recently @vks4git suggested that we merge this first to have this code in repo and not have it hang for long. I will create a separate issues for suggestions outlined above (FFA refactoring, Pairing refactoring, separate Symbolic datatype for Ed25519, anything else?)

Ok, agreed. @vks4git, could you resolve non-Plonk constraint, please? Do we need to merge the `Eq` instance in this PR?

vlasin (approved):

TurtlePU (approved):

PR #294, opened 2024-10-11
WIP list

vks4git:
My drafts for the Symbolic list. Feel free to discuss

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/List.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/List.hs
@@ -1,41 +1,158 @@
+{-# LANGUAGE TypeOperators #-}
+
 module ZkFold.Symbolic.Data.List where
 
-import           Data.Foldable              (Foldable (..))
-import           Data.Kind                  (Type)
-import           GHC.Generics               (Par1)
-import           Prelude                    (flip, undefined, (.))
+import           Data.Kind                        (Type)
+import           Data.Proxy                       (Proxy (..))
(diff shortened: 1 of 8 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/List.hs
```diff
@@ -1,42 +1,158 @@
...
+            \vHash vRepr (Par1 s) -> V.zipWithM (\h r -> newAssigned (\p -> p h - p r * p s)) vHash vRepr
 
-head :: List context x -> x
-head xs = let (x, _) = uncons xs in x
+        decSize :: context Par1
+        decSize = fromCircuitF lSize (\(Par1 s) -> Par1 <$> newAssigned (\p -> p s - one))
 
-tail :: List context x -> List context x
-tail xs = let (_, xs') = uncons xs in xs'
+{-- TODO: Foldable relies on Haskell types such as Bool (i.e. null = foldr (\_ _ -> False) True)
```
TurtlePU, line 144: I think it shouldn't be Haskell's `Foldable`, but rather some similar typeclass which uses IVC inside

TurtlePU:
If this compiles after fixing merge conflicts, I'd rather merge this version and then refactor it while moving in the direction I mentioned in #295. Also maybe we can abstract away the hash algorithm (and IVC device, too?) used via an additional type parameter for `List` datatype

TurtlePU:
(Very nice PR btw)

TurtlePU (approved):
Nice! :shipit:

vlasin (approved):

PR #299, opened 2024-10-14
SymbolicData Generics for finitary product types

echatav:
WIP Generics for SymbolicData


symbolic-base/src/ZkFold/Base/Data/Par1.hs
TurtlePU, line 1: Nice!

symbolic-base/src/ZkFold/Base/Protocol/Protostar/Fold.hs
```diff
@@ -138,14 +138,15 @@ ivcVerifierAc
     => Support y ~ Proxy ctx
     => Representable (Layout i)
     => Representable (Layout c)
-    => Representable (Layout f)
+    -- => Representable (Layout f)
```
TurtlePU, line 141: These are required to compile `ivcVerifier` into circuit, better leave them as is

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
```diff
@@ -152,3 +163,40 @@ instance SymbolicData f => SymbolicData (x -> f) where
...
+    grestore f = grestore (hmap fstP . f) :*: grestore (hmap sndP . f)
+
+instance GSymbolicData f => GSymbolicData (G.M1 i c f) where
+    type GContext (G.M1 i c f) = GContext f
+    type GSupport (G.M1 i c f) = GSupport f
+    type GLayout (G.M1 i c f) = GLayout f
+    gpieces (G.M1 a) = gpieces a
+    grestore f = G.M1 (grestore f)
+
+instance SymbolicData x => GSymbolicData (G.Rec0 x) where
```
TurtlePU, line 197: `Rec0` encodes recursion, and we currently cannot have recursive `SymbolicData` (for example, this instance would cause the typechecker to loop because of infinite `GLayout (G.Rec0 x)` type definition)

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
```diff
@@ -24,15 +25,39 @@ import           ZkFold.Base.Data.Vector          (Vector)
...
+
     type Support x :: Type
+    type Support x = GSupport (G.Rep x)
+
     type Layout x :: Type -> Type
+    type Layout x = GLayout (G.Rep x)
 
     -- | Returns the circuit that makes up `x`.
     pieces :: x -> Support x -> Context x (Layout x)
+    default pieces
```
TurtlePU, line 40: The idea was to have a single _instance_ of `SymbolicData` like this: `instance (G.Generic x, GSymbolicData (G.Rep x)) => SymbolicData x` in order to not have users write it down at all, but this also works I guess

TurtlePU:
🔥 🔥 🔥 
Keep it up!

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
echatav: This isn't quite right. It won't cause infinite loop; as you can see the type families are inductively smaller in all cases. The `Rec0` instance is required for the tuple instances. Without it the following tuple instance cannot be derived from the tuple's `Generic` instance & the `GSymbolicData` instance for `:*:`, since the `G.Rep (x,y)` involves `Rec0`.

```
instance
    ( SymbolicData x
    , SymbolicData y
    , HApplicative (Context x)
    , Context x ~ Context y
    , Support x ~ Support y
    ) => SymbolicData (x, y) where
```

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
TurtlePU: Sorry, in light of #308, I think better solution would be to go your way with usual Generic deriving

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
TurtlePU: Ok, but what happens if we, for example, try to derive a `SymbolicData` for a datatype like this:
```haskell
data Stream c = MkStream (Bool c) (Stream c)
```
What does the GHC tell us? Does it loop? Does it error out?

echatav:
Rather than a _general_ solution for `SymbolicData` instances, I think `Generic` deriving, on first approximation, is a feature to enable users to derive `SymbolicData` for their (finitary) _record_ or _product_ datatypes when the fields of their datatypes are `SymbolicData`. In particular, it is not initially intended for infinite list/stream types, nor for sum types.

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
echatav: It loops on infinite types. It will fail with a complaint about _reduction-depth_ and loop infinitely if you set `-freduction-depth=0`.

```
ghci> :{
ghci| data Stream c = MkStream (Bool c) (Stream c)
ghci|   deriving stock Generic
ghci|   deriving anyclass SymbolicData
ghci| :}

<interactive>:33:21: error:
    • Reduction stack overflow; size = 2
      When simplifying the following type:
        ZkFold.Symbolic.Data.Class.GContext
          (M1
             GHC.Generics.C
             (MetaCons "MkStream" PrefixI ghc-prim-0.10.0:GHC.Types.False)
             (S1
                (MetaSel
                   GHC.Maybe.Nothing
                   NoSourceUnpackedness
                   NoSourceStrictness
                   DecidedLazy)
                (Rec0 (Bool c))
              :*: S1
                    (MetaSel
                       GHC.Maybe.Nothing
                       NoSourceUnpackedness
                       NoSourceStrictness
                       DecidedLazy)
                    (Rec0 (Stream c))))
      Use -freduction-depth=0 to disable this check
      (any upper bound you could choose might fail unpredictably with
       minor updates to GHC, so disabling the check is recommended if
       you're sure that type checking should terminate)
    • When deriving the instance for (SymbolicData (Stream c))
```

TurtlePU:
Ok, let's fix conflicts, [reduce the diff](https://github.com/zkFold/zkfold-base/pull/299#discussion_r1800238978) and merge this.

symbolic-base/src/ZkFold/Base/Protocol/Protostar/Fold.hs
echatav: I reverted these comments and merged in changes from main. However, this code fails in a way that is a bit inscrutable to me, subject to whether or not a `4`-tuple has hand-rolled or generically-derived `SymbolicData` instance. Any insight would be appreciated.

TurtlePU: That's unfortunate, I think we need to reconstruct `t` from scratch for this to work
And let's wait for #325 to merge

echatav: refactored `t`

TurtlePU (approved):

vlasin (approved):

PR #300, opened 2024-10-14
feat: add quick testing scenario with flag

diS3e:
Added flag to disable all slow tests


echatav (changes requested):
Can you remove commented code and if possible, make the tests run with the fast flag turned on in CI?

TurtlePU (changes requested):
I'd prefer the division into slow/fast tests to be more granular: we'd still like to test at least some of the simpler properties of proof protocols in CI, for example.

To achieve this, I'd do it via command line arguments specified in a CI: we can name slow tests as "...something something... SLOW" and skip them with [`--skip=SLOW` Hspec argument](https://hspec.github.io/options.html#specifying-options-in-config-fi%E2%80%8C%E2%80%8Bles).

TurtlePU (approved):

echatav (approved):
lgtm

PR #319, opened 2024-10-29
Vks4git/309/constraint error

vks4git:


symbolic-base/src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
```diff
@@ -53,19 +54,19 @@ instance (Ord a, Arbitrary a, Binary a, KnownNat i) => Arbitrary (PlonkConstrain
 
 toPlonkConstraint :: forall a i . (Ord a, FiniteField a, KnownNat i) => Poly a (Var a (Vector i)) Natural -> PlonkConstraint i a
 toPlonkConstraint p =
-    let xs    = toList (variables p)
+    let xs    = Just <$> toList (variables p)
         perms = nubOrd $ map (take 3) $ permutations $ case length xs of
-            0 -> [ConstVar one, ConstVar one, ConstVar one]
```
TurtlePU, line 58: Is this change needed?

symbolic-base/test/Tests/Plonkup.hs
```diff
@@ -33,10 +36,23 @@ import           ZkFold.Base.Protocol.Plonkup.Utils                  (sortByList
...
 
--- TODO: uncomment after refactoring
--- propPlonkConstraintConversion :: (Eq a, FiniteField a) => PlonkConstraint 1 a -> Bool
--- propPlonkConstraintConversion p =
---     toPlonkConstraint (fromPlonkConstraint p) == p
+-- | Polynomial types and specific polynomials that were causing exceptions
+--
+problematicPolynomials :: (Ord a, FiniteField a) => [PM.Poly a (Var a (Vector 1)) Natural]
+problematicPolynomials =
+    [ var (ConstVar one)
```
TurtlePU, line 43: These look like a part of the `ConstVar` test

symbolic-base/src/ZkFold/Base/Protocol/Plonkup/PlonkConstraint.hs
vks4git: I do not see how else we can separate `ConstVar one` here as a placeholder for an impossible variable and `ConstVar one` as an input polynomial. `perms` will hold exactly the same value if the input polynomial does not have any variables and if it is `ConsyVar one`

TurtlePU: But do we need to separate a placeholder from an input polynomial?

TurtlePU (approved):

PR #325, opened 2024-10-31
Protostar fix

vlasin:
- Previously, circuit computations sneaked in into the prover side. Now, they are completely separated. We work with abstract field and group elements, as well as public input and message types.
- The `ivcVerifier`  can be arithmetized provided that the aforementioned types are `SymbolicInput`.
- I have not touched the tests in this PR, but I think the type constraints that we have now make sense and should be solvable.


symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
```diff
@@ -22,94 +23,75 @@ import           ZkFold.Base.Protocol.Protostar.CommitOpen   (CommitOpen (..), C
...
+    , MapInput (PU.PolyVec f deg) a ~ [PU.PolyVec f deg]
+    , MapMessage (PU.PolyVec f deg) a ~ PU.PolyVec f deg
+    ) => AccumulatorScheme pi f c m (FiatShamir f (CommitOpen m c a)) where
+  prover (FiatShamir (CommitOpen _ sps)) ck acc (InstanceProofPair pubi (NARKProof pi_x pi_w)) =
+        (Accumulator (AccumulatorInstance pi'' ci'' ri'' eCapital' mu') m_i'', pf)
       where
           -- Fig. 3, step 1
           r_i :: [f]
-          r_i = P.tail $ P.scanl (P.curry oracle) (oracle pubi) (zero : pi_x)
+          r_i = P.scanl (P.curry oracle) (oracle pubi) pi_x
```
vks4git, line 80: Why this change? I understand that these are random challenges, and it's probably unimportant how they are calculated, but the old implementation corresponds to the paper

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
```diff
@@ -22,94 +23,75 @@ import           ZkFold.Base.Protocol.Protostar.CommitOpen   (CommitOpen (..), C
...
 
           -- Fig. 3, step 2
 
           -- X + mu as a univariate polynomial
           polyMu :: PU.PolyVec f deg
-          polyMu = PU.polyVecLinear (acc^.x^.mu) one
+          polyMu = PU.polyVecLinear one (acc^.x^.mu)
```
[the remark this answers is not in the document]
vlasin: It changed in one of the updates. Would it be better to change the comment and naming of the variables?

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
vlasin: Hmm, I don't see how the previous version corresponds to the paper. On the first iteration, you were calculating `oracle (oracle pubi) zero` and using it in the calculation for $r_i$. In the paper, on the first iteration, you do `oracle (oracle pubi) c_1` where `c_1` is the commitment of the first message. But I agree that I should take the `tail` as before.

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
```diff
@@ -22,94 +23,75 @@ import           ZkFold.Base.Protocol.Protostar.CommitOpen   (CommitOpen (..), C
...
-    , LinearCombinationWith f (MapInput f a)
-    , MapMessage f a ~ m
+    , RandomOracle pi f         -- Random oracle for compressing public input
+    , RandomOracle c f          -- Random oracle ρ_NARK
+    , HomomorphicCommit f [f] c
+    , HomomorphicCommit f m c
+    , IsList pi
+    , Item pi ~ f
+    , IsList m
+    , Item m ~ f
```
vks4git, line 63: Is this true, though? Prover messages in the Special sound protocol for circuits are Maps. Moreover, they require a different approach to linear combinations than lists: absent keys are treated as zero coefficients. So, `linearCombination [('a', 1), ('c', 3)] [('b', 2), ('c', 4)]` should yield `[('a', X), ('b', 2), ('c', 3X + 4)]`. That's why I introduced the `LinearCombination` class

TurtlePU: Still, I think `LinearCombination a b` can be replaced with `Scale a b, AdditiveSemigroup b`

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
vlasin: Well, in my opinion, r0 := ρNARK(pi) in the paper corresponds to just `oracle pubi`. Either way, you need two oracles while in the paper they are denoted the same.

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
vlasin: > Is this true, though? Prover messages in the Special sound protocol for circuits are Maps. Moreover, they require a different approach to linear combinations than lists: absent keys are treated as zero coefficients. So, `linearCombination [('a', 1), ('c', 3)] [('b', 2), ('c', 4)]` should yield `[('a', X), ('b', 2), ('c', 3X + 4)]`. That's why I introduced the `LinearCombination` class

It does not matter the context. In my understanding of the paper, the assumption is that `pi` and `m` can be somehow broken down into field elements (of the same context). So, I think `IsList` fits very well here.

vlasin: For example, `pi` could be some tuple or vector of field elements. Obviously, there is a way to write it as a list and that's all we need. We just need to make sure that it implements `toList`.

vlasin: For `pi` this is not explicitly stated. But for messages, see page 39. Clearly, all messages are just vectors of `f` of the appropriate size. When we apply the "commit-open" and "Fiat-Shamir" transformations we only use tuple and list operators, so `toList` can still be easily defined for such messages. Actually, if we promote the number of rounds to the type level, then we only use the tuple operator.

symbolic-base/src/ZkFold/Base/Protocol/Protostar/AccumulatorScheme.hs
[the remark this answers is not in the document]
vlasin: I updated the comments.

TurtlePU (dismissed):

vks4git (approved):

vks4git:
:fire:

TurtlePU (approved):

TurtlePU:
Nice!

PR #327, opened 2024-11-02
Optimization of instance Extend (UInt n r c) (UInt k r c)

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -177,15 +177,41 @@ instance
...
+
+                resZ <- helper zs ks
+                let (_, res) = Haskell.unzip resZ
+                return $ V.unsafeToVector res
+
+            n = registerSize @(BaseField c) @n @r
+            k = registerSize @(BaseField c) @k @r
+
+            zm :: (MonadCircuit i (BaseField c) m) => m i
+            zm = newAssigned (const zero)
```
TurtlePU, line 197: we can now use `FromConstant (BaseField c) i` for this instead

TurtlePU (dismissed):

vlasin (approved):

PR #331, opened 2024-11-06
Refactor range constraints in ArithmeticCircuit

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -55,32 +56,33 @@ import           ZkFold.Base.Data.HFunctor
...
 type Constraint c i = Poly c (SysVar i) Natural
 
 -- | Arithmetic circuit in the form of a system of polynomial constraints.
-data ArithmeticCircuit a i o = ArithmeticCircuit
+data ArithmeticCircuit a (l :: [Lookup]) i o = ArithmeticCircuit
     {
         acSystem  :: Map ByteString (Constraint a i),
         -- ^ The system of polynomial constraints
-        acRange   :: Map (SysVar i) a,
+        acLookup  :: Map Lookup (S.Set (SysVar i)),
```
TurtlePU, line 70: This is incorrect definition because there may be (and will be) binary lookups, not only unary ones

TurtlePU:
For now, as for _refactoring range constraints_, I would just change the type of `acRange` to `Map a (Set (SysVar i))` as we don't know the proper abstraction for lookups yet
Maybe the proposed approach would work, but I don't like that this way the information about available lookups will bubble up to the user and they will need to specify it in order to use Symbolic

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -204,8 +205,8 @@ instance
       in
         zoom #acSystem . modify $ insert (toVar @a p) (p evalConstVar)
 
-    rangeConstraint (SysVar v) upperBound =
-      zoom #acRange . modify $ insert v upperBound
+    rangeConstraint (SysVar v) l =
+      zoom #acRange . modify $ insert l $ S.singleton v
```
TurtlePU, line 209: This would replace the accumulated set with a singleton, but we need to append to it

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -267,7 +268,7 @@ apply ::
   i a -> ArithmeticCircuit a (i :*: j) U1 -> ArithmeticCircuit a j U1
 apply xs ac = ac
   { acSystem = fmap (evalPolynomial evalMonomial varF) (acSystem ac)
-  , acRange = mapKeys' (acRange ac)
+  , acRange = S.fromList . catMaybes . toList . filterSet <$> acRange ac
```
TurtlePU, line 271: We can also check here if provided values satusfy the range constraints

TurtlePU (approved):

vlasin (approved):

Issue #307, opened 2024-10-15
Add support for extra witnesses in `Symbolic` and `MonadCircuit`


TurtlePU:
Symbolic `List`s require witnesses which should not correspond to any variable in a constraint system. We can achieve this by introducing a separate `WitnessField` type family in `Symbolic` class which can be used inside `MonadCircuit`:
```haskell
class ( FromConstant a w, FiniteField w, ToConstant w
      , FromConstant (Const w) w, SemiEuclidean (Const w)
      ) => Witness i a w | w -> i, w -> a where
  var :: i -> w

class (Witness var a w, Monad m) => MonadCircuit var a w m | m -> var, m -> a, m -> w where
  unconstrained :: w -> m var
  ...

class (HApplicative c, Package c, Arithmetic (BaseField c)) => Symbolic c where
  type BaseField c :: Type
  type WitnessField c :: Type

  witness :: c f -> f (WitnessField c)
  fromCircuitF ::
    c f -> (forall i m. MonadCircuit i (BaseField c) (WitnessField c) m => f i -> m (g i)) -> c g
  ...
```
Subtasks:
- [x] Add `Witness` typeclass and its instances: for current `WitnessField` type alias and for interpretation in `Interpreter`
- [x] Incorporate `Witness` into `MonadCircuit`, refactor its usages
- [ ] Change `acWitness` field type to allow for future `witness` implementation (note that this also unlocks serialization of the witness)
- [ ] Add `WitnessField` type family to `Symbolic`

PR #334, opened 2024-11-07
+ Witness class for use in MonadCircuit

TurtlePU:
Adds `Witness` class which is immediately used in `MonadCircuit` as described in #307, usages refactored.
Also some minor refactorings to improve readability along the way.

Added all of you to reviewers mostly to notify you because this is an important change in the `MonadCircuit` API


vlasin (approved):

hovanja2011 (dismissed):

symbolic-base/src/ZkFold/Symbolic/Data/ByteString.hs
```diff
@@ -168,15 +168,12 @@ instance (Symbolic c, KnownNat n) => BoolType (ByteString n c) where
...
-        where
-            solve :: MonadCircuit i (BaseField c) m => Vector n i -> m (Vector n i)
-            solve xv = do
-                let xs = V.fromVector xv
-                ys <-  for xs $ \i -> newAssigned (\p -> one - p i)
-                return $ V.unsafeToVector ys
+    not (ByteString bits) = ByteString $ fromCircuitF bits $ \xv -> do
+        let xs = V.fromVector xv
+        ys <- for xs $ \i -> newAssigned (\p -> one - p i)
+        return $ V.unsafeToVector ys
```
vks4git, line 174:
```suggestion
    not (ByteString bits) = ByteString $ fromCircuitF bits $ mapM (\i -> newAssigned (\p -> one - p i))
```
Vectors are Traversable, would this work?

vks4git (approved):

symbolic-base/src/ZkFold/Symbolic/Data/ByteString.hs
TurtlePU: Yes, this would, thank you!

Issue #310, opened 2024-10-27
Implement ledger rules of the Symbolic Ledger


vlasin:
See our internal PDF document.

I have already created a draft of ledger rules. The current task is to update and complete it in accordance with the docs.

The key difference from the draft is that "public" and "private" inputs are replaced with "online" and "offline" transactions. Online transactions allow us to achieve maximal data compression but require the user to be online at the moment when the rollup batch is submitted to the blockchain.

The other difference is that the bridge logic is streamlined: for bridged-in assets, we simply generate new UTxOs in the Symbolic ledger. For bridged-out assets, we send those to an unspendable address.

Sub-tasks:

- [x] #338
- [x] #339
- [x] #340
- [x] #341
- [x] #342
- [x] #343

PR #346, opened 2024-11-11
Implement ledger rules

echatav:
Address #310


zkfold-ledger/src/ZkFold/Symbolic/Ledger/Types/Update.hs
```diff
@@ -2,30 +2,33 @@ module ZkFold.Symbolic.Ledger.Types.Update where
...
 -- | Block hash of the corresponding block.
-type UpdateId context = Hash context
+type UpdateId = Hash
+
+type AddressIndex = UInt 40 Auto
 
--- | Update is a public data contained in a block.
--- TODO: This has been significantly redesigned. See the internal PDF docs.
--- Instead of "public inputs and outputs", we now have "online and offline transactions".
 data Update context = Update
```
vlasin, line 20: 1. The Merkle tree root for online transactions is missing. Online transactions require two-round communication between the address owner and the aggregation server in which the address owner gets the path to their leaf in the Merkle tree.
3. The list of released indices is missing. The idea is that if the money from a particular address is completely spent, then the address is removed from the index. That way, we are protected against index spam attacks.

echatav: 1. Added `updateId` as a record field. I had misunderstood that it was derivable somehow from the other fields.

2. Added `updateIndicesReleased` as a record field.

vlasin: `updateId` is not a good name for it because it may be confused with the hash of the `Update` type. This field is a Merkle tree root of the online transactions in the update.

echatav: changed the name to `updateOnlineTxsRoot`

vlasin (approved):

PR #361, opened 2024-11-14
Variables can be forcefully evaluated

vks4git:


symbolic-base/src/ZkFold/Symbolic/Compiler.hs
```diff
@@ -77,6 +78,7 @@ compileForceOne ::
     , Representable l
     , Binary (Rep l)
     , Ord (Rep l)
+    , NFData (Rep l)
```
TurtlePU, line 81: Can you also add `NfData (Rep (Layout x))` to superclass constraints of `SymbolicInput x` typeclass definition? This way we can get rid of some of constraints here

TurtlePU (approved):
Let's merge after adding `NFData` to `SymbolicInput`
LGTM!

vlasin (approved):

PR #371, opened 2024-11-20
Added `compileWithPayload` combinator

### LINKED ISSUE #370: A function that transforms (part of) inputs into extra witness

TurtlePU:
Resolves #370 by adding a new flavor of `compile` function which takes inputs from the payload instead of usual input variables

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Compiler.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Compiler.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeOperators       #-}
 
 module ZkFold.Symbolic.Compiler (
@@ -7,9 +6,13 @@ module ZkFold.Symbolic.Compiler (
     compile,
     compileIO,
     compileForceOne,
+    compileCircuit,
+    compileWithPayload,
+    solderWith,
     solder,
 ) where
 
+import           Control.DeepSeq                            (NFData)
 import           Data.Aeson                                 (FromJSON, ToJSON, ToJSONKey)
 import           Data.Binary                                (Binary)
 import           Data.Function                              (const, (.))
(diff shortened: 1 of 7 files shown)


symbolic-base/src/ZkFold/Symbolic/Compiler.hs
```diff
@@ -44,85 +47,86 @@ import           ZkFold.Symbolic.MonadCircuit               (MonadCircuit (..))
...
+  forall a p c f s l y .
+  ( c ~ ArithmeticCircuit a l p, Arithmetic a, Binary a
+  , Binary (Rep p), Ord (Rep p), NFData (Rep p)
+  , Traversable l
+  , SymbolicData f, Context f ~ c, Support f ~ s
+  , SymbolicInput s, Context s ~ c, Layout s ~ l
+  , SymbolicData y, Context y ~ c, Support y ~ Proxy c
+  , Layout f ~ Layout y
+  ) => f -> y
+compileWithPayload = compileCircuit . solderWith payloadCircuit
```
vlasin, line 108: Proposal: let's consider having a single `compile` function with an additional "options" argument. Would it help avoid the mistake of selecting a "wrong" compile function?

TurtlePU: Well, we can leave only the `compileWith f` function which would look like `compileCircuit . f`, where `f` can be a `solderWith c` postcomposed with some desired operations

TurtlePU: Left only `solderWith`, `finalRestore`, `compile` and `compileIO` in the public API

symbolic-base/src/ZkFold/Symbolic/Compiler.hs
```diff
@@ -41,88 +37,50 @@ import           ZkFold.Symbolic.MonadCircuit               (MonadCircuit (..))
...
-    , Layout f ~ Layout y
-    , Symbolic c
-    ) => f -> y
-compile = restore . const . optimize . solder @a
+compile :: forall a y f c s p.
+  ( SymbolicData f, Context f ~ c, Support f ~ s, SymbolicInput s, Context s ~ c
+  , c ~ ArithmeticCircuit a p (Layout s), Arithmetic a, Binary a, Binary (Rep p)
+  , SymbolicData y, Context y ~ c, Support y ~ Proxy c, Layout y ~ Layout f)
+  => f -> y
+compile = finalRestore . solderWith idCircuit
```
vlasin, line 62: I propose adding `compileWith` function that has parameters: a circuit for `solderWith` and a transformation after the `solderWith` function (like `forceOne` or `guessOutput`).

vlasin (approved):

PR #373, opened 2024-11-21
Generic SymbolicInput

### LINKED ISSUE #372: Default overridable implementation of `SymbolicInput` via `Generic`:

echatav:
resolves #372

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
@@ -15,7 +15,6 @@ import           GHC.TypeLits                       (KnownNat)
 import           Prelude                            (foldl, ($), (.))
 
 import           ZkFold.Base.Algebra.Basic.Class
-import           ZkFold.Base.Control.HApplicative   (HApplicative)
 import           ZkFold.Base.Data.ByteString        (Binary)
 import           ZkFold.Base.Data.Vector            (Vector, fromVector)
 import           ZkFold.Symbolic.Class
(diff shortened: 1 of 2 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
```diff
@@ -72,11 +75,31 @@ instance (
...
+
+class
+    ( GSymbolicData u
+    ) => GSymbolicInput u where
+    gisValid :: u x -> Bool (GContext u)
+
+instance
+    ( GContext u ~ GContext v
+    , GSupport u ~ GSupport v
+    , HApplicative (GContext v)
```
TurtlePU, line 94: `HApplicative` is implied by `Symbolic`

symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
TurtlePU, line 1: I think an `instance GSymbolicInput (K1 ...)` is missing, is it not?

symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
echatav: removed

symbolic-base/src/ZkFold/Symbolic/Data/Input.hs
echatav: There's an instance for the synonym `Rec0 = K1 R`, needed for default tuple instances. It can be generalized to `K1 i` since `i` is a phantom type anyway if you want it. I didn't do this for `SymbolicData` but could do it there as well.

TurtlePU: Oh, ok, just didn't see the loopback which goes like `gisValid (...) = isValid ...` so thought that maybe some instance was missing

TurtlePU (approved):

PR #374, opened 2024-11-21
Deeper constant optimization

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
```diff
@@ -33,46 +33,115 @@ module ZkFold.Symbolic.Compiler.ArithmeticCircuit (
...
+import           ZkFold.Symbolic.MonadCircuit                            (MonadCircuit (..))
 
 --------------------------------- High-level functions --------------------------------
 
 -- | Optimizes the constraint system.
 --
--- TODO: Implement nontrivial optimizations.
-optimize :: ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
-optimize = id
+optimize1 :: forall a p i o. ( Arithmetic a, Ord (Rep i))
```
vlasin, line 78: Please add the description of what it does.

TurtlePU: Also I don't think that the State monad is needed here, it can be expressed as just `SysVar i -> a -> AC a p i o -> AC a p i o`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
```diff
@@ -33,46 +33,115 @@ module ZkFold.Symbolic.Compiler.ArithmeticCircuit (
...
+      WSysVar nV@(NewVar _) ->
+        if v == nV
+        then ww (WSysVar nV)
+        else fromConstant k
+      x -> ww x
+
+optimize ::
+  ( Arithmetic a, Ord (Rep i)) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize = until (not . haveOptimal) f
```
TurtlePU, line 116: I think we can avoid a `haveOptimal` call by matching on the emptiness of `[toConstVar p | ...]`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
TurtlePU, line 1: I think optimization deserves its own module, `ZkFold.Symbolic.Compiler.ArithmeticCircuit.Optimization`

vlasin (approved):

symbolic-examples/stats/Conditional
```diff
@@ -1,4 +1,4 @@
 Conditional
 Number of constraints: 10
-Number of variables: 9
+Number of variables: 5
```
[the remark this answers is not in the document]
TurtlePU: Well, actually this optimization triggers quite rare by itself, only on constraints which are added on `NewVar`s (like in usage of `isValid`, for example)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,93 @@
...
+
+optimize1 :: forall a p i o. ( Arithmetic a, Ord (Rep i))
+  => (SysVar i , a )-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimize1 (v, k) ac = case v of
+  NewVar nv ->  ac {
+      acSystem = M.filter (/= zero) (M.map optPoly $ acSystem ac),
+      acRange =  MM.filter (/= S.empty) $ optRanges $ acRange ac,
+      acWitness = (>>= optWitVar) <$> M.delete nv (acWitness ac)
+      }
+  _ -> error "This shouldn't happen"
```
TurtlePU, line 47: While this doesn't happen with circuits in oractice, it's still theoretically possible: it means that our system constraints one of inputs to be constant.

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,93 @@
...
+  ( Arithmetic a, Ord (Rep i)) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = case [toConstVar p | (_, p) <- toList (acSystem ac), isLinear p] of
+    [] -> ac
+    vs -> optimize $ foldr optimize1 ac vs
+
+optimize1 :: forall a p i o. ( Arithmetic a, Ord (Rep i))
+  => (SysVar i , a )-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimize1 (v, k) ac = case v of
+  NewVar nv ->  ac {
```
TurtlePU, line 42: `acOutput` might change as well if one of the outputs turns out to be a constant

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,93 @@
...
+    optWitVar :: WitVar p i -> WitnessF a (WitVar p i)
+    optWitVar = \case
+      (WSysVar (NewVar nV)) ->
+        if NewVar nV == v
+          then WitnessF $ const $ fromConstant k
+          else pure $ WSysVar (NewVar nV)
+      w  -> pure w
+
+
+toConstVar :: (Arithmetic a, Ord (Rep i)) => Constraint a i -> (SysVar i, a)
```
TurtlePU, line 73: The `toConstVar`/`isLinear` split can be avoided if we replace it with a single function of type `Constraint a i -> Maybe (SysVar i, a)`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,93 @@
...
+
+-- | Replaces linear polynoms of the form
+-- @(fromConstant k) * (NewVar nV) + (fromConstant c)@
+-- with a constant variable @fromConstant $ negate c // k@ in an arithmetic circuit
+-- and replaces variable with a constant in witness
+--
+optimize ::
+  ( Arithmetic a, Ord (Rep i)) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = case [toConstVar p | (_, p) <- toList (acSystem ac), isLinear p] of
```
TurtlePU, line 35: Current performance is ok I guess, but if we need it to be faster later, we can do the following:
1. Notice that the optimization passes depend only on the state of `acSystem`
2. Instead of updating the whole circuit on each pass, we can only update the constraint system, accumulating the list of constant variables
3. After the system stabilizes (no more constants can be inferred), we can apply the whole accumulated list of constant variables to the rest of the circuit instead of applying it one-by-one

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
vlasin: Yes. It should not change the circuit, since we cannot change the public inputs.

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+      acOutput = acOutput ac <&> \case
+        SysVar nV@(NewVar _) -> if nV == v then ConstVar k else SysVar nV
+        o -> o
+      }
+  _ -> ac
+  where
+    optRanges :: MM.MonoidalMap a (S.Set (SysVar i)) -> MM.MonoidalMap a (S.Set (SysVar i))
+    optRanges = MM.mapWithKey (\r s -> bool s (newS r s) (S.member v s) )
+      where
+        newS r s = bool (error "range constraint less then value") (S.filter (/= v) s) (k <= r)
```
TurtlePU, line 75: `S.delete` can be used instead of `S.filter`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+  (Arithmetic a, Ord (Rep i), Functor o) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = let (ac', vs) = varsToReplace (ac, []) in foldr optimizeAc ac' vs
+
+varsToReplace ::
+  (Arithmetic a, Ord (Rep i), Functor o) =>
+  (ArithmeticCircuit a p i o , [(SysVar i, a)]) -> (ArithmeticCircuit a p i o , [(SysVar i, a)])
+varsToReplace (ac, l) = case catMaybes [toConstVar p | (_, p) <- toList (acSystem ac)] of
+    [] -> (ac, l)
+    vs -> varsToReplace (ac {acSystem = acSystem $ foldr optimizeSystems ac vs}, l ++ vs)
```
TurtlePU, line 43: It's faster to prepend `vs` instead of appending: `vs ++ l`

TurtlePU: Also it's faster to substitute several variables at once instead of one-by-one

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+
+-- | Replaces linear polynoms of the form
+-- @(fromConstant k) * (NewVar nV) + (fromConstant c)@
+-- with a constant variable @fromConstant $ negate c // k@ in an arithmetic circuit
+-- and replaces variable with a constant in witness
+--
+optimize ::
+  (Arithmetic a, Ord (Rep i), Functor o) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = let (ac', vs) = varsToReplace (ac, []) in foldr optimizeAc ac' vs
```
TurtlePU, line 36: Same as with system optimization: it's faster to substitute variables in bulk instead of one-by-one

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+
+varsToReplace ::
+  (Arithmetic a, Ord (Rep i), Functor o) =>
+  (ArithmeticCircuit a p i o , [(SysVar i, a)]) -> (ArithmeticCircuit a p i o , [(SysVar i, a)])
+varsToReplace (ac, l) = case catMaybes [toConstVar p | (_, p) <- toList (acSystem ac)] of
+    [] -> (ac, l)
+    vs -> varsToReplace (ac {acSystem = acSystem $ foldr optimizeSystems ac vs}, l ++ vs)
+
+optimizeSystems :: forall a p i o. (Arithmetic a, Ord (Rep i)) =>
```
TurtlePU, line 45: Since you're only modifying the system, it would be cleaner to only use a system as input/output instead of the whole circuit

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+  (SysVar i , a ) -> ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimizeSystems (v, k) ac = case v of
+  NewVar _ -> ac {acSystem = M.filter (/= zero) (M.map optPoly $ acSystem ac)}
+  _        -> ac
+  where
+    optMono :: (a, Mono (SysVar i) Natural) -> (a, Mono (SysVar i) Natural)
```
TurtlePU, line 51: In order to not have to define `optMono`/`optPoly`, I'd do the substitution via `evalPolynomial`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,99 @@
...
+
+optimizeAc :: forall a p i o. (Arithmetic a, Ord (Rep i), Functor o)
+  => (SysVar i , a ) -> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimizeAc (v, k) ac = case v of
+  NewVar nv -> ac {
+      acRange =  MM.filter (/= S.empty) $ optRanges $ acRange ac,
+      acWitness = (>>= optWitVar) <$> M.delete nv (acWitness ac),
+      acOutput = acOutput ac <&> \case
+        SysVar nV@(NewVar _) -> if nV == v then ConstVar k else SysVar nV
+        o -> o
```
TurtlePU, line 68: Again, it's perfectly fine to substitute `InVar`s for their values here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
TurtlePU: (Also maybe we can use `Map (SysVar i) a` instead of list to accumulate results)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,102 @@
...
+--
+optimize ::
+   (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i)) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = let (newAcSystem, vs) = varsToReplace (acSystem ac, M.empty) in optimizeAc vs ac {acSystem = newAcSystem}
+
+varsToReplace ::
+  (Arithmetic a, Ord (Rep i)) =>
+  (Map ByteString (Constraint a i) , Map (SysVar i) a) -> (Map ByteString (Constraint a i) , Map (SysVar i) a)
+varsToReplace (s, l) = let newVars = M.fromList . map fromJust . M.elems $ M.filter isJust $ M.map toConstVar s in
```
TurtlePU, line 45: I'd say `M.fromList $ mapMaybe toConstVar (M.elems s)`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,102 @@
...
+  then (s, l)
+  else varsToReplace (optimizeSystems newVars s, M.union newVars l)
+
+optimizeSystems :: (Arithmetic a, Ord (Rep i)) =>
+  Map (SysVar i) a -> Map ByteString (Constraint a i) -> Map ByteString (Constraint a i)
+optimizeSystems m as = M.filter (/= zero) (M.map deleteZeroM $ evalPolynomial evalMonomial varF <$> as)
+  where
+    varF p = maybe (var p) fromConstant (M.lookup p m)
+
+    deleteZeroM :: (Polynomial c (SysVar i) Natural) => Poly c (SysVar i) Natural -> Poly c (SysVar i) Natural
```
TurtlePU, line 56: I think it should be an invariant of polynomials thelat all coefficients are nonzero. If that's not the case, I think polynomials have to be fixed

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -86,7 +86,7 @@ instance (Arbitrary c, Arbitrary (Mono i j)) => Arbitrary (Poly c i j) where
 instance {-# OVERLAPPING #-} FromConstant (Poly c i j) (Poly c i j)
 
 instance Polynomial c i j => AdditiveSemigroup (Poly c i j) where
-    P l + P r = P $ go l r
+    P l + P r = P $ filter ((/= zero) . fst) $ go l r
```
TurtlePU, line 89: I think we need to do the same in multiplication

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+      acRange =  MM.filter (/= S.empty) $ optRanges vs (acRange ac),
+      acWitness = (>>= optWitVar vs) <$> M.filterWithKey (\k _ -> notMember (NewVar k) vs) (acWitness ac),
+      acOutput = acOutput ac <&> \case
+        SysVar sV -> maybe (SysVar sV) ConstVar (M.lookup sV vs)
+        o -> o
+      }
+  where
+    addInVarConstraints :: Map ByteString (Poly a (SysVar i) Natural) -> Map ByteString (Poly a (SysVar i) Natural)
+    addInVarConstraints p = foldrWithKey (\k a as -> case k of
+      inVar@(InVar inV) -> M.insert (runHash @(Just (Order a)) $ merkleHash inV) (var inVar - fromConstant a) as
```
TurtlePU, line 69: it's better to stay consistent with `constraint` implementation in `ArithmeticCircuit.Internal` when computing the key

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+      acSystem = addInVarConstraints $ acSystem ac,
+      acRange =  MM.filter (/= S.empty) $ optRanges vs (acRange ac),
+      acWitness = (>>= optWitVar vs) <$> M.filterWithKey (\k _ -> notMember (NewVar k) vs) (acWitness ac),
+      acOutput = acOutput ac <&> \case
+        SysVar sV -> maybe (SysVar sV) ConstVar (M.lookup sV vs)
+        o -> o
+      }
+  where
+    addInVarConstraints :: Map ByteString (Poly a (SysVar i) Natural) -> Map ByteString (Poly a (SysVar i) Natural)
+    addInVarConstraints p = foldrWithKey (\k a as -> case k of
```
TurtlePU, line 68: This can be done with `p <> fromList [ (toVar polyId, poly) | (InVar i, v) <- assocs vs, poly = var i - fromConstant v, polyId = ...]`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+  where
+    varF p = maybe (var p) fromConstant (M.lookup p m)
+
+optimizeAc :: forall a p i o. (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i))
+  => Map (SysVar i) a-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimizeAc vs ac =
+    ac {
+      acSystem = addInVarConstraints $ acSystem ac,
+      acRange =  MM.filter (/= S.empty) $ optRanges vs (acRange ac),
+      acWitness = (>>= optWitVar vs) <$> M.filterWithKey (\k _ -> notMember (NewVar k) vs) (acWitness ac),
```
TurtlePU, line 61: `filterWithKey` can be replaced with `\\`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+        o -> o
+      }
+  where
+    addInVarConstraints :: Map ByteString (Poly a (SysVar i) Natural) -> Map ByteString (Poly a (SysVar i) Natural)
+    addInVarConstraints p = foldrWithKey (\k a as -> case k of
+      inVar@(InVar inV) -> M.insert (runHash @(Just (Order a)) $ merkleHash inV) (var inVar - fromConstant a) as
+      _                 -> as) p vs
+
+    optRanges :: Map (SysVar i) a -> MM.MonoidalMap a (S.Set (SysVar i)) -> MM.MonoidalMap a (S.Set (SysVar i))
+    optRanges m = MM.mapWithKey (\r s -> bool (error "range constraint less then value") (S.difference s $ keysSet m) (all (<= r) $ restrictKeys m s))
```
TurtlePU, line 73: I think `mapMaybeWithKey` might be a better fit here (helps to get rid of `filter` afterwards)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+
+optimizeSystems :: (Arithmetic a, Ord (Rep i)) =>
+  Map (SysVar i) a -> Map ByteString (Constraint a i) -> Map ByteString (Constraint a i)
+optimizeSystems m as = M.filter (/= zero) $ evalPolynomial evalMonomial varF <$> as
+  where
+    varF p = maybe (var p) fromConstant (M.lookup p m)
+
+optimizeAc :: forall a p i o. (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i))
+  => Map (SysVar i) a-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimizeAc vs ac =
```
TurtlePU, line 57: I think we can inline a call to this function as it is not used anywhere else

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+  (Arithmetic a, Ord (Rep i)) =>
+  (Map ByteString (Constraint a i) , Map (SysVar i) a) -> (Map ByteString (Constraint a i) , Map (SysVar i) a)
+varsToReplace (s, l) = let newVars = M.fromList . M.elems $ mapMaybe toConstVar s in
+  if newVars == M.empty
+  then (s, l)
+  else varsToReplace (optimizeSystems newVars s, M.union newVars l)
+
+optimizeSystems :: (Arithmetic a, Ord (Rep i)) =>
+  Map (SysVar i) a -> Map ByteString (Constraint a i) -> Map ByteString (Constraint a i)
+optimizeSystems m as = M.filter (/= zero) $ evalPolynomial evalMonomial varF <$> as
```
TurtlePU, line 51: Same as range checks, maybe we want to throw an error here if we encounter an unsatisfiable constraint?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
TurtlePU: Also I'd try to inline functions as much as possible and write them as short as possible to make everything what is done as clear as possible

symbolic-base/test/Tests/Arithmetization/Optimization.hs
```diff
@@ -0,0 +1,71 @@
...
+
+testBool :: forall a . (Arithmetic a, Binary a) => ArithmeticCircuit a (U1 :*: U1) (Par1 :*: U1) Par1
+testBool = ac
+    where
+        identBool :: Bool c -> Bool c
+        identBool x = x
+
+        Bool ac = compile @a identBool
+
+testConst :: forall a . (Arithmetic a, Binary a) => ArithmeticCircuit a U1 Par1 Par1
```
TurtlePU, line 40: Isn't it the same as `testFunc`?

symbolic-base/test/Tests/Arithmetization/Optimization.hs
```diff
@@ -0,0 +1,71 @@
...
+    constraint (($ i4) - fromConstant (4 :: Natural))
+    return (Par1 i5)
+
+testBool :: forall a . (Arithmetic a, Binary a) => ArithmeticCircuit a (U1 :*: U1) (Par1 :*: U1) Par1
+testBool = ac
+    where
+        identBool :: Bool c -> Bool c
+        identBool x = x
+
+        Bool ac = compile @a identBool
```
TurtlePU, line 38: You can just `testBool = compile @a identBool`, no need to wrap it in `Bool` and unwrap back

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+      _                 -> as) p vs
+
+    optRanges :: Map (SysVar i) a -> MM.MonoidalMap a (S.Set (SysVar i)) -> MM.MonoidalMap a (S.Set (SysVar i))
+    optRanges m = MM.mapWithKey (\r s -> bool (error "range constraint less then value") (S.difference s $ keysSet m) (all (<= r) $ restrictKeys m s))
+
+    optWitVar :: Map (SysVar i) a -> WitVar p i -> WitnessF a (WitVar p i)
+    optWitVar m = \case
+      (WSysVar sv) ->
+        case M.lookup sv m of
+          Just k  -> WitnessF $ const $ fromConstant k
```
TurtlePU, line 79: Maybe just `fromConstant k` without `WitnessF $ const $` would do?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+--
+optimize ::
+   (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i)) =>
+  ArithmeticCircuit a p i o  -> ArithmeticCircuit a p i o
+optimize ac = let (newAcSystem, vs) = varsToReplace (acSystem ac, M.empty) in optimizeAc vs ac {acSystem = newAcSystem}
+
+varsToReplace ::
+  (Arithmetic a, Ord (Rep i)) =>
+  (Map ByteString (Constraint a i) , Map (SysVar i) a) -> (Map ByteString (Constraint a i) , Map (SysVar i) a)
+varsToReplace (s, l) = let newVars = M.fromList . M.elems $ mapMaybe toConstVar s in
```
TurtlePU, line 44: The shortest description of this algorithm would reveal itself when we define a combinator
```haskell
farthestM :: Monad m => (a -> Maybe (m a)) -> a -> m a
farthestM f = go where go x = maybe (return x) (>>= go) (f x)
```
And use it with `m ~ (, Map (SysVar i) a)` (and `farthestM` can be added to `ZkFold.Prelude`)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+optimizeAc :: forall a p i o. (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i))
+  => Map (SysVar i) a-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimizeAc vs ac =
+    ac {
+      acSystem = addInVarConstraints $ acSystem ac,
+      acRange =  MM.filter (/= S.empty) $ optRanges vs (acRange ac),
+      acWitness = (>>= optWitVar vs) <$> M.filterWithKey (\k _ -> notMember (NewVar k) vs) (acWitness ac),
+      acOutput = acOutput ac <&> \case
+        SysVar sV -> maybe (SysVar sV) ConstVar (M.lookup sV vs)
+        o -> o
```
TurtlePU, line 64: For documentation purposes, we can match on `ConstVar cV` instead of just `o` here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,98 @@
...
+optimizeSystems :: (Arithmetic a, Ord (Rep i)) =>
+  Map (SysVar i) a -> Map ByteString (Constraint a i) -> Map ByteString (Constraint a i)
+optimizeSystems m as = M.filter (/= zero) $ evalPolynomial evalMonomial varF <$> as
+  where
+    varF p = maybe (var p) fromConstant (M.lookup p m)
+
+optimizeAc :: forall a p i o. (Arithmetic a, Ord (Rep i), Functor o, Binary (Rep i))
+  => Map (SysVar i) a-> ArithmeticCircuit a p i o -> ArithmeticCircuit a p i o
+optimizeAc vs ac =
+    ac {
```
TurtlePU, line 58: As we change every field, it would be better to call an `ArithmeticCircuit` constructor here (maybe there will be other fields in circuit in future? we don't want to forget them too)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -0,0 +1,91 @@
...
+    acOutput = o <&> \case
+      SysVar sV -> maybe (SysVar sV) ConstVar (M.lookup sV vs)
+      so -> so}
+  where
+    (newS, vs) = varsToReplace (s, M.empty)
+
+    addInVarConstraints :: Map ByteString (Poly a (SysVar i) Natural) -> Map ByteString (Poly a (SysVar i) Natural)
+    addInVarConstraints p = p <> fromList [(polyId, poly) | (InVar i, v) <- assocs vs,
+                                                            let poly = var (InVar i) - fromConstant v,
+                                                            let polyId = toVar @a @p @i (pure $ WSysVar (InVar i))]
```
TurtlePU, line 49: In `Internal`, constraint id is computed on the whole constraint, not on the variable!

TurtlePU (approved):

TurtlePU (approved):

vlasin (approved):

PR #376, opened 2024-11-24
Added `SymbolicData.Payload` associated type

TurtlePU:
This PR resolves #375 by adding `Payload` type family to `SymbolicData` class. Consequently, this changes the interface a bit and some code around circuits/symbolic datatypes, so I invited everyone to take a look.

PR #380, opened 2024-11-25
Added `instance SymbolicData (List c x)`

### LINKED ISSUE #295: Add separate "witness layout" for `List`-like datatypes
Symbolic `List`, as designed, requires a separate list (Haskell's `[]`) of witnesses to function properly. The problem is, we cannot write a good `SymbolicData` instance for `List` as defined currently; however, this can be solved by the following additions:
```diff
class (HApplicative c, Package c) => Symbolic c where
  type BaseField c :: Type
+ type WitnessField c :: Type -- for circuits, this is Witness type from MonadCircuit

+ witness :: c f -> f (WitnessField c)
  ...

class SymbolicData d where
  type Context d :: (Type -> Type) -> Type
  type Support d :: Type
  type Layout d :: Type -> Type -- old layout functor
+ type Payload d :: Type -> Type -- new witness functor

- pieces :: d -> Support d -> Context d (Layout d)
+ pieces :: Context d ~ c => d -> Support d -> (c (Layout d), Payload d (Witness c))
- restore :: (Support d -> Context d (Layout d)) -> d
+ restore :: Context d ~ c => (Support d -> (c (Layout d), Payload d (Witness c))) -> d

-data ArithmeticCircuit a i o = ArithmeticCircuit {
+data ArithmeticCircuit a p i o = ArithmeticCircuit {
  acSystem :: Map ByteString (Constraint a i),
  acRange :: Map (SysVar i) a,
- acWitness :: Map ByteString (i a -> Map ByteString a -> a),
+ acWitness :: Map ByteString (p a -> i a -> Map ByteString a -> a),
  acOutput :: o (Var a i)
  }
```
Setting `Payload` for lists of type `x` to be a `Stream :.: (Layout x :*: Payload x)` (where `Stream` is an infinite list with `Rep Stream = Natural`), we get a `SymbolicData` instance for `List`s.
Note that the signature for `compile` function stays (almost) the same: we only need to provide a correct parameter for `p`, which is `Payload (Support f)`.

Subtasks:
- [x] Resolve #307
- [x] #363
- [x] #375
- [x] #379

### LINKED ISSUE #379: Implement `SymbolicData` for `List`

TurtlePU:
Resolves #379. This finally resolves #295.
Added the long awaited instance. Also introduced some helper orphan instances along the way.
(Note: #376 has to be merged first)


symbolic-base/src/ZkFold/Symbolic/Data/List.hs
```diff
@@ -1,45 +1,75 @@
...
+data List c x = List
+  { lHash    :: c (Layout x)
+  , lSize    :: c Par1
+  , lWitness :: Payloaded (Infinite :.: ListItem x) c
+  }
+  deriving (Generic)
+
+instance HApplicative c => SymbolicData (List c x)
+-- | TODO: Maybe some 'isValid' check for Lists?..
+instance (Symbolic c, SymbolicInput x) => SymbolicInput (List c x)
```
vlasin, line 61: Good question! On the one hand, we could check the "knowledge" of the list in `isValid`. That way, we can be sure that the prover has full knowledge of the arguments that they are passing to the computation. On the other hand, I cannot currently construct an example where it would cause a problem.

vlasin: The same applies to the `Hash` types.

vlasin:
Can we have some tests involving `head` and `tail` operations for Symbolic lists?

symbolic-base/src/ZkFold/Symbolic/Data/List.hs
TurtlePU: Well, yeah, we could define
```haskell
isValid (l :: List c x) = foldr (.:) emptyList l == l
isValid (h :: Hash c x) = hash (preimage h) == h
```
But in the end, this would introduce practically the same constraints as the ones user is going to introduce by using the operations on the datatype anyway, so I don't know if this is useful.

vlasin: Ok, let's leave those notes for now and revisit it later.

echatav:
Infinite lists are also isomorphic to [`Cofree Identity`](https://hackage.haskell.org/package/free/docs/Control-Comonad-Cofree.html).

echatav (approved):
LGTM

PR #384, opened 2024-11-26
Generic Symbolic Eq

### LINKED ISSUE #385: Default overridable implementation of `Eq` via `Generic`:

echatav:
Allow default implementations for our `Eq` for product and record types.
Resolves #385


symbolic-base/src/ZkFold/Symbolic/Data/Eq.hs
TurtlePU, line 1: I'd also add the `Eq (Bool c) (c f)` and `Eq (Bool c) (Bool c)` instance here and get rid of `Structural` newtype altogether

echatav: `Structural` equality and `Lexicographical` ordering are slightly more general than generic equality and ordering. I was thinking of unifying them to a single `newtype` but keeping it to allow `deriving via`s in some situation where the default generic may not work. What do you think?

TurtlePU: > What do you think?

I think that actually Generic deriving would yield more correct instances. Consider the following datatype:

```haskell
data Maybe c x = Maybe { isJust :: Bool c, fromJust :: x }
```

While we could derive via `Structural`, this would be incorrect because:
1. `Eq (Bool c) (Bool c)` is more efficient than `Eq (Bool c) (Structural (Bool c))`
2. `Eq (Bool c) (Structural x)` is simply unrelated to `Eq (Bool c) x` and would work incorrectly

And for every datatype where we could use `Structural`, deriving via Generic would also work.

So, all in all, deriving via Generic would yield more correct and more efficient instances.

symbolic-base/src/ZkFold/Symbolic/Data/Eq.hs
```diff
@@ -62,3 +69,24 @@ instance (Symbolic c, Haskell.Eq (BaseField c), Representable f, Traversable f)
         in
             any Bool (unpacked result)
 
+instance Symbolic c => Eq (Bool c) (Bool c)
```
TurtlePU, line 72: `Eq (Bool c) (Bool c)` can be implemented in a single constraint instead of two in a Generic implementation

TurtlePU: Ok, let's have it in a separate PR

TurtlePU (approved):

PR #392, opened 2024-12-01
Non-zero constraint in UInt tests

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -389,7 +389,7 @@ instance
                 (zs, b) <- flip runStateT b0 $ traverse StateT (Haskell.zipWith (fullSub r) is js)
                 d <- newAssigned (\v -> v i' - v j')
                 s'0 <- newAssigned (\v -> v d + v b + fromConstant t)
-                (s', _) <- splitExpansion (highRegisterSize @(BaseField c) @n @r) 1 s'0
+                (s', _) <- splitExpansion (highRegisterSize @(BaseField c) @n @r) (r + 1 -! highRegisterSize @(BaseField c) @n @r) s'0
```
TurtlePU, line 392: Nice catch! 
Though I think it's better to fix the definition of `s'0` to use `2 ^ highRegisterSize - 1` instead of `t` in order to have less constraints

TurtlePU (approved):

vlasin (approved):

PR #394, opened 2024-12-03
Refactored Plonk(up) to use arbitrary functors

### LINKED ISSUE #386: Refactor Plonk / PlonkUp to use structured inputs

TurtlePU:
Resolves #386.
As said in the title: refactored Plonk(up) to use arbitrary functors instead of Vectors.

### DIFF
--- a/symbolic-base/src/ZkFold/Base/Protocol/Plonkup/Internal.hs
+++ b/symbolic-base/src/ZkFold/Base/Protocol/Plonkup/Internal.hs
@@ -3,9 +3,9 @@
 
 module ZkFold.Base.Protocol.Plonkup.Internal where
 
-import           Data.Binary                                         (Binary)
 import           Data.Constraint                                     (withDict)
 import           Data.Constraint.Nat                                 (plusNat, timesNat)
+import           Data.Functor.Classes                                (Show1)
 import           Data.Functor.Rep                                    (Rep)
 import           Prelude                                             hiding (Num (..), drop, length, sum, take, (!!),
                                                                       (/), (^))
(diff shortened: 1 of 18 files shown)


symbolic-base/src/ZkFold/Base/Protocol/Plonkup/Internal.hs
```diff
@@ -24,11 +23,11 @@ import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.Internal
     Additionally, we don't want this library to depend on Cardano libraries.
 -}
 
-data Plonkup p (i :: Natural) (n :: Natural) (l :: Natural) curve1 curve2 transcript = Plonkup {
+data Plonkup p i (n :: Natural) l curve1 curve2 transcript = Plonkup {
```
vlasin, line 26: I am wondering if it makes sense to remove `n` as a type parameter. We cannot know the number of constraints in advance, so it might cause more confusion than good (as experienced, for instance, by @diS3e).

TurtlePU: Well, it makes sense to use it in some of the datatypes in Plonk, but there it can be introduced using [`withDict`](https://hackage.haskell.org/package/constraints-0.14.2/docs/Data-Constraint.html#v:withDict) with the exact number of constraints in the circuit

TurtlePU: So yes, in most cases it can be removed

vlasin (approved):

PR #396, opened 2024-12-10
EC symbolic

echatav:
Change the `Point` type and friends so that they may support Symbolic computation.

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/Bool.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/Bool.hs
@@ -22,7 +22,6 @@ import           Text.Show                       (Show)
 
 import           ZkFold.Base.Algebra.Basic.Class
 import           ZkFold.Symbolic.Class
-import           ZkFold.Symbolic.Data.Class      (SymbolicData)
 import           ZkFold.Symbolic.Interpreter     (Interpreter (..))
 import           ZkFold.Symbolic.MonadCircuit    (newAssigned)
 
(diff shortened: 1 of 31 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/Bool.hs
```diff
@@ -65,8 +64,6 @@ deriving instance Show (c Par1) => Show (Bool c)
 instance {-# OVERLAPPING #-} (Eq a, MultiplicativeMonoid a) => Show (Bool (Interpreter a)) where
     show (fromBool -> x) = if x == one then "True" else "False"
 
-deriving newtype instance Symbolic c => SymbolicData (Bool c)
```
TurtlePU, line 68: Why did this instance have to be moved?

TurtlePU:
Really like the way elliptic curve API changed! So clean and polymorphic now

Hope we will sort the stuff with Conditional/Eq/Ord out so that modules in `Base` do not depend on `Symbolic`, just rubs me the wrong way...

symbolic-base/src/ZkFold/Symbolic/Data/Bool.hs
echatav: Hmmm, it might not have to have been. I had a cycle at one point but not sure if this in fact solved it or something else.

TurtlePU:
(Though I still believe that we should have Symbolic-specific versions of Eq and Ord classes for a cleaner API for Symbolic users which are not involved in language-polymorphic stuff)

echatav:
^ I think this is best handled by re-exporting and properly documenting in friendly ways these terms and others in a `Symbolic.Prelude`.

TurtlePU:
Well, I'm only concerned about Symbolic functions which are polymorphic in datatypes. Take an example like this:
```haskell
find :: a -> Vector n a -> Maybe c (UInt k r c)
```
What should the constraints look like?

With language-agnostic Eq
```haskell
find :: (Symbolic c, Eq a, BooleanOf a ~ Bool c) => a -> Vector n a -> Maybe c (UInt k r c)
```

With Symbolic-specific Eq
```haskell
class SymbolicData a => Eq a where
  (==) :: a -> a -> Bool (Context a)

find :: (Eq a, Context a ~ c) => a -> Vector n a -> Maybe c (UInt k r c)
```

P.S. We can introduce type aliases however we want, sure, but the real interface the user is going to interact with are compiler errors, and usually type aliases are not present there

PR #399, opened 2024-12-11
Linear optimization

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Class.hs
TurtlePU, line 1: Good idea to move this to a separate module! I'd call it `Var` instead of `Class`, though
And I don't know if `UVar` belongs here because it is not really a variable type IMO

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Class.hs
```diff
@@ -0,0 +1,140 @@
...
+instance FromConstant a (Var a i) where
+    fromConstant = ConstVar
+
+data UVar a i
+  = ConstUVar a
+  | LinUVar a (SysVar i) a
+  | More
+  deriving Generic
+
+deriving anyclass instance (FromJSON (Rep i), FromJSON a) => FromJSON (UVar a i)
```
TurtlePU, line 61: This pile of instances is not needed for `UVar` as the type is only used for witness estimation

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Class.hs
```diff
@@ -0,0 +1,140 @@
...
+deriving stock instance Eq (Rep i) => Eq (SysVar i)
+deriving stock instance Ord (Rep i) => Ord (SysVar i)
+deriving instance NFData (Rep i) => NFData (SysVar i)
+
+data Var a i
+  = LinVar a (SysVar i) a
+  | ConstVar a
+  deriving Generic
+
+toLinVar :: (Arithmetic a) => SysVar i -> Var a i
```
TurtlePU, line 41: I'd call this just `toVar`, the thing that it is linear is not essential to the caller imo

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Class.hs
```diff
@@ -0,0 +1,140 @@
...
+deriving anyclass instance (ToJSON (Rep i), ToJSON a) => ToJSONKey (Var a i)
+deriving anyclass instance (ToJSON (Rep i), ToJSON a) => ToJSON (Var a i)
+deriving stock instance (Show (Rep i), Show a) => Show (Var a i)
+deriving stock instance (Eq (Rep i), Eq a) => Eq (Var a i)
+deriving stock instance (Ord (Rep i), Ord a) => Ord (Var a i)
+deriving instance (NFData (Rep i), NFData a) => NFData (Var a i)
+instance FromConstant a (Var a i) where
+    fromConstant = ConstVar
+
+data UVar a i
```
TurtlePU, line 55: I think that this is not really a variable so it deserves its own module, maybe `WitnessEstimation` or something like this

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -127,31 +115,16 @@ pmapWitVar ::
...
-deriving anyclass instance (ToJSON (Rep i), ToJSON a) => ToJSONKey (Var a i)
-deriving anyclass instance (ToJSON (Rep i), ToJSON a) => ToJSON (Var a i)
-deriving stock instance (Show (Rep i), Show a) => Show (Var a i)
-deriving stock instance (Eq (Rep i), Eq a) => Eq (Var a i)
-deriving stock instance (Ord (Rep i), Ord a) => Ord (Var a i)
-deriving instance (NFData (Rep i), NFData a) => NFData (Var a i)
-instance FromConstant a (Var a i) where
-    fromConstant = ConstVar
-
 imapVar ::
```
TurtlePU, line 118: This is a part of `Var`'s interface so I'd move it there

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -92,18 +92,6 @@ instance (NFData a, NFData (o (Var a i)), NFData (Rep i))
...
-deriving anyclass instance FromJSON (Rep i) => FromJSON (SysVar i)
-deriving anyclass instance FromJSON (Rep i) => FromJSONKey (SysVar i)
-deriving anyclass instance ToJSON (Rep i) => ToJSONKey (SysVar i)
-deriving anyclass instance ToJSON (Rep i) => ToJSON (SysVar i)
-deriving stock instance Show (Rep i) => Show (SysVar i)
-deriving stock instance Eq (Rep i) => Eq (SysVar i)
-deriving stock instance Ord (Rep i) => Ord (SysVar i)
-deriving instance NFData (Rep i) => NFData (SysVar i)
 
 imapSysVar ::
```
TurtlePU, line 96: This is a part of `SysVar`'s interface so I'd move it there

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -211,43 +184,47 @@ instance
...
-        Just c -> if c ==zero
+        Just c -> if c == zero
                     then return ()
                     else error "The constraint is non-zero"
         Nothing -> zoom #acSystem . modify $ insert (toVar @_ @p (p at)) (p evalConstVar)
 
-    rangeConstraint (SysVar v) upperBound =
-      zoom #acRange . modify $ insertWith S.union upperBound (S.singleton v)
+    rangeConstraint (LinVar k x b) upperBound =
+      zoom #acRange . modify $ insertWith S.union ((upperBound - b) // k) (S.singleton x)
```
TurtlePU, line 227: Oops! Looks like we completely wrecked the range constraints here
In finite fields, range constraint "y <= b" actually means "y \in {0, 1,...,b}" because there's no proper ordering actually
So "kx+b<=c" actually means "x \in {-b/k, (1-b)/k,..., (c-b)/k}"
Note that there's no guarantee at all that this is a contiguous interval from 0 to (c-b)/k
I think there's two ways to remedy this:
1. Give up and create a new variable `y:=kx+b` and range constrain it in case k and b mess things up (pretty much always except for either (k=1, b=0) or (k=-1, b=c))
2. Generalize range constraints to lookup constraints of arbitrary subsets of a field
Pinging @vlasin because he might find this interesting

TurtlePU: (@hovanja2011, thank you for finding this! I definitely missed this when thinking things out)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Optimization.hs
```diff
@@ -38,7 +38,7 @@ optimize (ArithmeticCircuit s r w o) = ArithmeticCircuit {
     acRange = optRanges vs r,
     acWitness = (>>= optWitVar vs) <$>  M.filterWithKey (\k _ -> notMember (NewVar k) vs) w,
     acOutput = o <&> \case
-      SysVar sV -> maybe (SysVar sV) ConstVar (M.lookup sV vs)
+      lv@(LinVar k sV b) -> maybe lv (ConstVar . (\t -> k * t + b)) (M.lookup sV vs)
```
TurtlePU, line 41: Actually, being a constant-or-linear expression, `Var` is a monad with `>>=` being a substitution (same as with `WitnessF`). Using this fact, the code here can be simplified to `o <&> (>>= \v -> maybe (pure v) ConstVar (M.lookup v vs))`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+instance (MultiplicativeSemigroup a, Eq a, AdditiveMonoid a) => Scale a (UVar a i) where
+  scale k (ConstUVar c) = ConstUVar $ k * c
+  scale k (LinUVar a x b) = if k == zero
+    then ConstUVar zero
+    else LinUVar (k * a) x (k * b)
+  scale k More = if k == zero
+    then ConstUVar zero
+    else More
+
+instance Scale Natural (UVar a i) where scale k = scale k . fromConstant
```
TurtlePU, line 41: `fromConstant` ends up being applied to the `UVar` itself which would cause this function to loop. I think you meant `scale k = scale (fromConstant k :: a)`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  | LinUVar a (SysVar i) a
+  | More
+  deriving Generic
+
+instance FromConstant a (UVar a i) where
+    fromConstant = ConstUVar
+instance FromConstant Natural a => FromConstant Natural (UVar a i) where fromConstant = ConstUVar . fromConstant
+instance FromConstant Integer a => FromConstant Integer (UVar a i) where fromConstant = ConstUVar . fromConstant
+
+instance (MultiplicativeSemigroup a, Eq a, AdditiveMonoid a) => Scale a (UVar a i) where
```
TurtlePU, line 32: I think distributivity is also required, so `a` is at least a `Semiring`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+    then ConstUVar zero
+    else LinUVar (k * a) x (k * b)
+  scale k More = if k == zero
+    then ConstUVar zero
+    else More
+
+instance Scale Natural (UVar a i) where scale k = scale k . fromConstant
+instance Scale Integer (UVar a i) where scale k = scale k . fromConstant
+
+instance (Exponent a Natural, MultiplicativeMonoid a) => Exponent (UVar a i) Natural where
```
TurtlePU, line 44: `Exponent a Natural` is implied by `MultiplicativeMonoid a`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  (LinUVar {}) ^ 0    = ConstUVar one
+  _ ^ _               = More
+
+instance (Exponent a Integer, MultiplicativeMonoid a) => Exponent (UVar a i) Integer where
+  (ConstUVar c) ^ n   = ConstUVar $ c ^ n
+  (LinUVar k x b) ^ 1 = LinUVar k x b
+  _ ^ 0               = ConstUVar one
+  _ ^ _               = More
+
+instance (AdditiveSemigroup a, AdditiveMonoid a, Eq a, Eq (Rep i)) => AdditiveSemigroup (UVar a i) where
```
TurtlePU, line 56: `AdditiveMonoid` implies `AdditiveSemigroup`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+      else LinUVar (k1 + k2) x1 (b1 + b2)
+    else More
+  LinUVar k x b + ConstUVar c = LinUVar k x (b + c)
+  ConstUVar c + LinUVar k x b = LinUVar k x (b + c)
+  _ + _ = More
+
+instance (AdditiveMonoid a, Eq a, Eq (Rep i)) => AdditiveMonoid (UVar a i) where
+  zero = ConstUVar zero
+
+instance (Eq a, AdditiveMonoid a, Eq (Rep i), AdditiveGroup a) => AdditiveGroup (UVar a i) where
```
TurtlePU, line 70: `AdditiveGroup` implies `AdditiveMonoid`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  negate (ConstUVar c)   = ConstUVar (negate c)
+  negate (LinUVar k x b) = LinUVar (negate k) x (negate b)
+  negate More            = More
+
+instance (MultiplicativeSemigroup a, Eq a, AdditiveMonoid a, Eq (Rep i)) => MultiplicativeSemigroup (UVar a i) where
```
TurtlePU, line 75: I'd say that distributivity is expected so `Semiring a` is a better fit instead of `MultiplicativeSemigroup` and `AdditiveMonoid`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  ConstUVar c1 * ConstUVar c2 = ConstUVar $ c1 * c2
+  ConstUVar c * (LinUVar k x b) = if c == zero
+    then zero
+    else LinUVar (c * k) x (c * b)
+  (LinUVar k x b) * (ConstUVar c) = if c == zero
+    then zero
+    else LinUVar (c * k) x (c * b)
+  _ * _ = More
```
TurtlePU, line 83: `ConstUVar 0 * More = ConstUVar 0`, I think

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+
+instance (MultiplicativeMonoid a, AdditiveMonoid a, Eq a, Eq (Rep i)) => MultiplicativeMonoid (UVar a i) where
```
TurtlePU, line 85: `Semiring` implies both monoids

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  one = ConstUVar one
+
+instance (AdditiveMonoid a, Eq a, Eq (Rep i), FromConstant Natural a, MultiplicativeMonoid a) => Semiring (UVar a i)
```
TurtlePU, line 88: `Semiring` implies both monoids and `FromConstant Natural`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+
+instance (AdditiveMonoid a, Eq a, Eq (Rep i), AdditiveGroup a, FromConstant Natural a, FromConstant Integer a, MultiplicativeMonoid a) => Ring (UVar a i)
```
TurtlePU, line 90: `Ring a` implies everything except for `Eq a` and `Eq (Rep i)` and would be a better fit here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+
+instance (AdditiveMonoid a, Eq a, Eq (Rep i), AdditiveGroup a, FromConstant Natural a, FromConstant Integer a, MultiplicativeMonoid a, Field a) => Field (UVar a i) where
```
TurtlePU, line 92: `Field a` implies everything except `Eq a` and `Eq (Rep i)` here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,101 @@
...
+  finv (ConstUVar c) = ConstUVar $ finv c
+  finv _             = More
+
+instance (ToConstant a) => ToConstant (UVar a i) where
+  type Const (UVar a i) = Const a
+  toConstant (ConstUVar c) = toConstant c
+  toConstant _             = error "WitnessEstimation: this should`t happen"
```
TurtlePU, line 99: This happens though! Look at the code in `instance ToConstant WitnessF` for insights on what to do here

symbolic-base/test/Tests/Arithmetization/Optimization.hs
```diff
@@ -53,7 +53,7 @@ specOptimization' = do
         it "number of constraint decreases" $ do
             acSizeN ac === 1
         it "number of variables decreases" $ do
-            acSizeM ac === 1
+            acSizeM ac === 0
```
TurtlePU, line 56: So actually `optimize` does nothing here because of `LinVars`
Maybe add a single multiplication at the start of `testFunc`/`testConst` to stop the linear optimization from stealing all the work from `optimize`?

TurtlePU: (I'd add this as another circuit and another bunch of tests instead of replacing the original tests, maybe)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -211,43 +172,55 @@ instance
     type WitnessField (ArithmeticCircuit a p i) = WitnessF a (WitVar p i)
     witnessF (behead -> (c, o)) = o <&> \case
```
TurtlePU, line 173: With optimization you introduced, it is possible to just use `at` here instead of pattern-matching

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -211,43 +172,55 @@ instance
...
+        Just c -> if c == zero
                     then return ()
                     else error "The constraint is non-zero"
-        Nothing -> zoom #acSystem . modify $ insert (toVar @_ @p (p at)) (p evalConstVar)
-
-    rangeConstraint (SysVar v) upperBound =
-      zoom #acRange . modify $ insertWith S.union upperBound (S.singleton v)
+        Nothing -> zoom #acSystem . modify $ insert (witToVar @_ @p (p at)) (p evalConstVar)
+
+    rangeConstraint (LinVar k x b) upperBound = if k == one && b == zero
```
TurtlePU, line 214: I'd extract the operation of "prepare `SysVar` which is suitable for range constraint" in a separate function.
Also the case `k == -1 && b == upperBound` is also ok here.

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+  finv _             = More
+
+instance ToConstant a => ToConstant (UVar a i) where
+  type Const (UVar a i) = Maybe (Const a)
+  toConstant (ConstUVar c) = Just $ toConstant c
+  toConstant _             = Nothing
+
+instance Finite a => Finite (UVar a i) where type Order (UVar a i) = Order a
+
+instance (SemiEuclidean a, Eq a, Eq (Rep i)) => SemiEuclidean (UVar a i) where
```
TurtlePU, line 105: Since we use `Maybe` for `ToConstant`, this instance is not needed anymore

symbolic-examples/stats/UInt.DivMod.32.Auto
```diff
@@ -1,4 +1,4 @@
 UInt.DivMod.32.Auto
-Number of constraints: 1387
-Number of variables: 1735
-Number of range lookups: 766
\ No newline at end of file
+Number of constraints: 1292
+Number of variables: 1642
+Number of range lookups: 770
```
TurtlePU, line 4: Hmm, it's interesting that the number of ranged constraints slightly increased in comparison to the code before optimizations

symbolic-examples/stats/SHA256.32
```diff
@@ -1,4 +1,4 @@
 SHA256.32
-Number of constraints: 50566
-Number of variables: 60683
-Number of range lookups: 10996
\ No newline at end of file
+Number of constraints: 51184
```
TurtlePU, line 2: I feel like we forgot to run benchmarks on some previous commit which caused the regression, can you check the result of this benchmark on `main` branch?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+  div More _                        = More
+  div _ More                        = ConstUVar zero
+  mod (ConstUVar c1) (ConstUVar c2) = ConstUVar $ mod c1 c2
+  mod (ConstUVar c) _               = ConstUVar c
+  mod (LinUVar _ _ b) (ConstUVar c) = ConstUVar $ mod b c
+  mod (LinUVar {}) (LinUVar {})     = More
+  mod (LinUVar k x b) More          = LinUVar k x b
+  mod More _                        = More
+
+instance (FromConstant Natural a) => FromConstant (Maybe Natural) (UVar a i) where
```
TurtlePU, line 119: Do we need this instance?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+  ConstUVar c * (LinUVar k x b) = if c == zero
+    then ConstUVar zero
+    else LinUVar (c * k) x (c * b)
+  ConstUVar c * More = if c == zero
+    then ConstUVar zero
+    else More
+  (LinUVar k x b) * (ConstUVar c) = if c == zero
+    then ConstUVar zero
+    else LinUVar (c * k) x (c * b)
+  _ * _ = More
```
TurtlePU, line 85: Also `More * ConstUVar 0 = ConstUVar 0`

TurtlePU: You can reduce the number of cases here by doing:
```haskell
ConstUVar c * x = scale c x
x * ConstUVar c = scale c x
_ * _ = More
```

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+
+instance (AdditiveMonoid a, Eq a, Eq (Rep i)) => AdditiveSemigroup (UVar a i) where
+  ConstUVar c1 + ConstUVar c2 = ConstUVar $ c1 + c2
+  LinUVar k1 x1 b1 + (LinUVar k2 x2 b2) = if x1 == x2
+    then if k1 + k2 == zero
+      then ConstUVar $ b1 + b2
+      else LinUVar (k1 + k2) x1 (b1 + b2)
+    else More
+  LinUVar k x b + ConstUVar c = LinUVar k x (b + c)
+  _ + _ = More
```
TurtlePU, line 64: `ConstUVar + LinUVar` case is missing

TurtlePU: I'd recommend doin the same as in multiplication:
```haskell
ConstUVar c + x = c .+ x
x + ConstUVar c = c .+ x
LinUVar + LinUVar = ... 
_ + _ = More
```

TurtlePU: (Where `.+` is defined separately)

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+    then ConstUVar zero
+    else More
+
+instance (Semiring a, Eq a) => Scale Natural (UVar a i) where scale k = scale (fromConstant k :: a)
+instance Eq a => Scale Integer (UVar a i) where scale k = scale k . fromConstant
+
+instance MultiplicativeMonoid a => Exponent (UVar a i) Natural where
+  (ConstUVar c) ^ n   = ConstUVar $ c ^ n
+  (LinUVar k x b) ^ 1 = LinUVar k x b
+  (LinUVar {}) ^ 0    = ConstUVar one
```
TurtlePU, line 47: More ^ 0 is also a 1

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+      then ConstUVar $ b1 + b2
+      else LinUVar (k1 + k2) x1 (b1 + b2)
+    else More
+  LinUVar k x b + ConstUVar c = LinUVar k x (b + c)
+  _ + _ = More
+
+instance (Semiring a, Eq a, Eq (Rep i)) => AdditiveMonoid (UVar a i) where
+  zero = ConstUVar zero
+
+instance (AdditiveGroup a, Semiring a, Eq a, Eq (Rep i)) => AdditiveGroup (UVar a i) where
```
TurtlePU, line 69: `AdditiveGroup` and `Semiring` imply `Ring`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+    else LinUVar (c * k) x (c * b)
+  ConstUVar c * More = if c == zero
+    then ConstUVar zero
+    else More
+  (LinUVar k x b) * (ConstUVar c) = if c == zero
+    then ConstUVar zero
+    else LinUVar (c * k) x (c * b)
+  _ * _ = More
+
+instance (MultiplicativeMonoid a, Eq a, Semiring a) => MultiplicativeMonoid (UVar a i) where
```
TurtlePU, line 87: `MultiplicativeMonoid` is implied by `Semiring`

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/WitnessEstimation.hs
```diff
@@ -0,0 +1,121 @@
...
+  one = ConstUVar one
+
+instance (Semiring a, Eq a, Eq (Rep i), MultiplicativeMonoid a) => Semiring (UVar a i)
```
TurtlePU, line 90: MultiplicativeMonoid is implied by Semiring

symbolic-examples/stats/SHA256.32
hovanja2011: The benchmark result on the `main` branch is indeed smaller. Somehow, transformations increase the number of range constraints.

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -210,44 +171,59 @@ instance
...
-        let v = toVar @a wf
-        -- TODO: forbid reassignment of variables
-        zoom #acWitness $ modify (insert v wf)
-        return $ SysVar (NewVar v)
+    unconstrained wf =
+        case runWitnessF wf $ \case
+          WSysVar sV -> LinUVar one sV zero
+          _          -> More
+        of
+          LinUVar k x b -> return (LinVar k x b)
```
TurtlePU, line 193: Forgot `ConstUVar` case here! That's why range constraints reappeared

TurtlePU: Here's lesson for the future: do not use wildcards in pattern matching

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
TurtlePU: I'm talking about just `at <$> o`, so `\case` is not needed here at all

TurtlePU (approved):

TurtlePU (approved):

vlasin (approved):

PR #406, opened 2024-12-16
fix `ListItem` definition

### LINKED ISSUE #401: Fix `ListItem` definition
When run in circuit context, the `Payload` argument for `ListItem` ends up containing the whole circuit type, leading to the recursive definition of input payload, which fails to typecheck.

We need to change the definition of `ListItem` so that stops being the case and circuit compiles fine.

TurtlePU:
Resolves #401

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/List.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/List.hs
@@ -29,30 +29,22 @@ import           ZkFold.Symbolic.Data.Payloaded   (Payloaded (Payloaded))
 import           ZkFold.Symbolic.Data.UInt        (UInt)
 import           ZkFold.Symbolic.MonadCircuit
 
-data ListItem x a = ListItem
-  { tailHash    :: Layout x a
-  , headLayout  :: Layout x a
-  , headPayload :: Payload x a
+data ListItem l p a = ListItem
+  { tailHash    :: l a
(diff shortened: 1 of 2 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/List.hs
```diff
@@ -29,30 +29,22 @@ import           ZkFold.Symbolic.Data.Payloaded   (Payloaded (Payloaded))
...
-    Distributive (ListItem x) where
+instance (Representable l, Representable p) => Distributive (ListItem l p) where
   distribute = distributeRep
 
-instance
-  (Representable (Layout x), Representable (Payload x)) =>
-  Representable (ListItem x)
+instance (Representable l, Representable p) => Representable (ListItem l p)
 
 data List c x = List
```
echatav, line 44: Is there any reason we need to parameterize `List` by its context `c` since it can be determined from `x` as `Context x` just as `Layout x` & `Payload x` are?

TurtlePU: A good question, actually
I think we can, same in `Maybe`
We can create an issue abt this

TurtlePU: #415

vlasin (approved):

PR #407, opened 2024-12-17
Custom `UInt` layout

### LINKED ISSUE #297: Create custom `Layout` for `UInt` to avoid excessive type-level computations
Currently, a `Layout (UInt n r c)` is a `Vector (NumberOfRegisters (BaseField c) n r)`, which is a can of worms:
1. To work with low and high registers, we need to push/pop them from the vector, which is suboptimal;
2. Most operations end up requiring `KnownNat (NumberOfRegisters (BaseField c) n r)` which clutter the constraints;
3. All in all, we need to compute `NumberOfRegisters (BaseField c) n r`, which leads to slow compilation and unreadable compilation errors.

All of this can be solved with a custom `Layout (UInt n r c)` with a custom `Representable` instance which would only require `Symbolic c`, `KnownNat n` and `KnownRegisterSize r`, which is much more tractable.

P.S. Note that we do not need to make a custom `SymbolicData` instance for `UInt`, it's the functor which is used inside that has to be changed from `Vector` to something else.

TurtlePU:
Resolves #297.

To simplify constraints and to speed up compilation (Haskell part) a bit, I added a custom layout for `UInt` which doesn't require `KnownNat (NumberOfRegisters ...)`

### DIFF
--- a/symbolic-base/test/Tests/UInt.hs
+++ b/symbolic-base/test/Tests/UInt.hs
@@ -2,8 +2,6 @@
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeOperators       #-}
 
-{-# OPTIONS_GHC -freduction-depth=0 #-} -- Avoid reduction overflow error caused by NumberOfRegisters
-
 module Tests.UInt (specUInt) where
 
 import           Control.Applicative                         ((<*>))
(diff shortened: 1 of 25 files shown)


symbolic-base/test/Tests/UInt.hs
TurtlePU, line 1: Regression in test runtime, investigating

TurtlePU:
See discussion in #297.

PR #410, opened 2024-12-17
IVC update 1

vlasin:
This PR implements full Protostar IVC except for the following extra features:

- Compressed verifier transform
- Lookup and circuit selector sub-protocols

The IVC tests are also missing as we are waiting for the FFA update.


symbolic-base/src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -38,6 +39,11 @@ class FromConstant a b where
 
 instance FromConstant a a
 
+instance {-# OVERLAPPING #-} FromConstant (Constant a f) (Constant a f)
```
TurtlePU, line 42: Maybe we can collect algebraic instances for `Constant` together below, how it's done with `(e -> a)` and standard numeric types?

symbolic-base/src/ZkFold/Base/Data/Orphans.hs
TurtlePU, line 1: So do we intend to move here all orphan instances for types and classes which are defined in other packages?

vlasin: I am open to suggestions. I noticed that we have some in other modules.

symbolic-base/src/ZkFold/Base/Protocol/IVC/FiatShamir.hs
```diff
@@ -0,0 +1,48 @@
...
+    ) => f -> Vector k (c f) -> Vector (k-1) f
+transcript r0 cs = withDict (plusMinusInverse1 @1 @k) $ init $ init $ scanl (curry (oracle @algo)) r0 cs
+
+fiatShamir :: forall algo k i p c m o f .
+    ( KnownNat k
+    , HashAlgorithm algo f
+    , RandomOracle algo f f
+    , RandomOracle algo (i f) f
+    , RandomOracle algo (c f) f
+    ) => CommitOpen k i p c m o f -> FiatShamir k i p c m o f
```
TurtlePU, line 32: Can the type signature be generalized here to obtain a $1$-round SPS from any $k$-round SPS, not only from commit-open scheme?

symbolic-base/src/ZkFold/Base/Protocol/IVC/VerifierCircuit.hs
```diff
@@ -0,0 +1,70 @@
...
+-- import           ZkFold.Symbolic.Compiler
+-- import           ZkFold.Symbolic.Data.Class          (SymbolicData (..))
+-- import           ZkFold.Symbolic.Data.FieldElement   (FieldElement (..))
+-- import           ZkFold.Symbolic.Data.Input          (SymbolicInput)
+-- import ZkFold.Base.Protocol.IVC.SpecialSound (SpecialSoundProtocol(..))
+
+-- -- | Takes a function `f` and returns a circuit `C` with input `y` and witness `w`.
+-- -- The circuit is such that `C(y, w) = 0` implies that `y = x(n)` for some positive `n` where
+-- -- `x(k+1) = f(x(k), u(k))` for all `k` and some `u`.
+-- ivcVerifierCircuit :: forall f i p m c d k a payload input output nx nu ctx algo .
```
TurtlePU, line 28: So this would be re-added in some other way, right?

symbolic-base/src/ZkFold/Base/Protocol/Protostar.hs
TurtlePU, line 1: Why did the name change from Protostar to just IVC, by the way?

vlasin: Because the plan is to modify it next according to the ProtoGalaxy paper. I am considering potential modifications from other papers as well, but they have certain trade-offs. Anyway, I don't want to change module names every time.

symbolic-base/src/ZkFold/Base/Protocol/IVC/VerifierCircuit.hs
vlasin: Yes

symbolic-base/src/ZkFold/Base/Protocol/IVC/FiatShamir.hs
vlasin: Yes, this should work, thanks! I will try to include it in the next IVC PR.

PR #421, opened 2024-12-24
logic hierarchy

echatav:


symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,220 @@
...
+  ) => Conditional (x0,x1,x2,x3)
+
+instance (R.Representable v, Conditional x)
+  => Conditional (G.Generically1 v x) where
+    type BooleanOf (G.Generically1 v x) = BooleanOf x
+    bool (G.Generically1 vf) (G.Generically1 vt) b =
+      G.Generically1 (R.mzipWithRep (\f t -> bool f t b) vf vt)
+
+deriving via G.Generically1 (Vector n) x
+  instance (KnownNat n, Conditional x) => Conditional (Vector n x)
```
echatav, line 105: The `Conditional` (as well as `Eq`) instance for `Vector`s is derived via the `Generically1` newtype. This is because the instance definition works for any `Representable` (and `Foldable`) over a `Conditional` (and `Eq`). But the choice of the `Generically1` newtype is arbitrary. Similarly the choice of the `Identity` newtype for deriving `Conditional` (and `Eq`) instances from their Prelude variants is arbitrary.

Are their better newtypes to use for this purpose?

symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,220 @@
...
+import qualified Data.Functor.Rep                 as R
+import           Data.Kind
+import qualified GHC.Generics                     as G
+import           Prelude                          (type (~))
+import qualified Prelude                          as H
+
+import           ZkFold.Base.Algebra.Basic.Number
+import           ZkFold.Base.Data.Vector
+
+class Boolean b where
```
TurtlePU, line 29: I think some documentation describing the laws of the class could be useful here

symbolic-base/src/ZkFold/Base/Data/Logical.hs
TurtlePU: I think `Generically1` is fine

symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,220 @@
...
+      G.Generically1 (R.mzipWithRep (\f t -> bool f t b) vf vt)
+
+deriving via G.Generically1 (Vector n) x
+  instance (KnownNat n, Conditional x) => Conditional (Vector n x)
+
+instance Conditional (Identity x) where
+  type BooleanOf (Identity x) = H.Bool
+  bool = H.bool
+
+ifThenElse :: Conditional a => BooleanOf a -> a -> a -> a
```
TurtlePU, line 111: Maybe it is worth making this a class member and remove `bool` altogether? Trying to remember which arguments are which in `bool` were always hard to remember for me, for example

symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,220 @@
...
+
+instance Boolean H.Bool where
+  true = H.True
+  false = H.False
+  not = H.not
+  (&&) = (H.&&)
+  (||) = (H.||)
+  xor = (H./=)
+
+all :: (Boolean b, Foldable t) => (x -> b) -> t x -> b
```
TurtlePU, line 45: I think it would be better to remove `BoolType` and things which depend on it in this same PR to avoid dead code and code duplication in the repo

symbolic-base/src/ZkFold/Base/Data/Logical.hs
TurtlePU, line 1: I think `Boolean`, `Conditional` and `Eq` should each reside in their own module, namely `ZkFold.Base.Data.{Boolean, Conditional, Eq}`

symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,220 @@
...
+
+  (==) :: a -> a -> BooleanOf a
+  default (==) ::
+    ( G.Generic a
+    , GEq (G.Rep a)
+    , BooleanOf a ~ GBooleanOf (G.Rep a)
+    ) => a -> a -> BooleanOf a
+  x == y = geq (G.from x) (G.from y)
+
+  (/=) :: a -> a -> BooleanOf a
```
TurtlePU, line 124: I wonder if `(/=)` can be defined as `x /= y = not (x == y)` as, thanks to recent optimizations by @hovanja2011, this should have the same number of constraints in circuits

symbolic-base/src/ZkFold/Base/Data/Logical.hs
TurtlePU: Maybe it's better to use something else instead of `Identity` for Haskell types though...

symbolic-base/src/ZkFold/Base/Data/Logical.hs
```diff
@@ -0,0 +1,216 @@
...
+  instance (KnownNat n, Eq x) => Eq (Vector n x)
+
+instance H.Eq x => Eq (Identity x) where
+  (==) = (H.==)
+  (/=) = (H./=)
+
+elem :: (Eq a, Foldable t) => a -> t a -> BooleanOf a
+elem x = any (== x)
+
+class Boolean (GBooleanOf f) => GConditional f where
```
TurtlePU, line 168: Suggestion: move deriving stuff to corresponding `.Generic` modules (e.g. `ZkFold.Base.Data.Eq.Generic`) to make modules shorter and more readable. I guess instances for `K1` would still need to be defined in the main module, but this still would look cleaner and understandable (because it would be near instances of the main class for tuples which need it)

echatav: So 5 modules instead of 1? 🙃 I don't understand how this would work and not require mutually importing modules. I split it into 3 for now.

TurtlePU: ok, maybe at least visually separate the part on Generic deriving from the rest by a comment

symbolic-base/src/ZkFold/Base/Data/Eq.hs
```diff
@@ -0,0 +1,96 @@
...
+
+  (==) :: a -> a -> BooleanOf a
+  default (==) ::
+    ( G.Generic a
+    , GEq (G.Rep a)
+    , BooleanOf a ~ GBooleanOf (G.Rep a)
+    ) => a -> a -> BooleanOf a
+  x == y = geq (G.from x) (G.from y)
+
+  (/=) :: a -> a -> BooleanOf a
```
TurtlePU, line 32: Well, if `(/=)` is defined like this, then the definition from `GEq` is not used at all
Sorry for the previous comment, let's keep `(/=)` as you defined it for now

symbolic-base/src/ZkFold/Base/Data/Conditional.hs
```diff
@@ -0,0 +1,100 @@
...
+  ) => Conditional (x0,x1,x2,x3)
+
+instance (R.Representable v, Conditional x)
+  => Conditional (G.Generically1 v x) where
+    type BooleanOf (G.Generically1 v x) = BooleanOf x
+    ifThenElse b (G.Generically1 vt) (G.Generically1 vf) =
+      G.Generically1 (R.mzipWithRep (ifThenElse b) vt vf)
+
+deriving via G.Generically1 (Vector n) x
+  instance (KnownNat n, Conditional x) => Conditional (Vector n x)
```
TurtlePU, line 65: If we intend to use this instance in a generic context, then it is not as efficient as can be in a usual Haskell context as this would require $O(n)$ instead of the standard $O(1)$

TurtlePU: Don't know what to do here right now, I have the following idea but I don't know if this looks clean to you
```haskell
class Boolean b where
  ...
  choice :: (a -> a -> b -> a) -> (e -> a) -> (e -> a) -> b -> (e -> a)
  -- chooses a closure for @b ~ base.Data.Bool@, creates new closure for @b ~ ZkFold.Symbolic.Data.Bool@
```

TurtlePU: (Terminology borrowed from [here](https://github.com/JetBrains/arend-lib/blob/master/src/Logic/Classical.ard#L12) as this looks familiar)

TurtlePU: OR we could define two overlapping instances for `Vector`, but this would require the user to propagate the `Conditional (Vector n x)` constraint until `BooleanOf x` is known to be either `base.Data.Bool` or not

TurtlePU: OR we could define this `choice` to have a type `(Conditional a, BooleanOf a ~ b) => Choose (e -> a) b`, but this is not really better looking and makes `Boolean` and `Conditional` depend on each other unless we drop the dependency on `Boolean` in `Conditional`, but I don't like this approach

symbolic-base/src/ZkFold/Base/Data/Logical.hs
echatav: This is a draft PR while I’m working on it. Don’t worry I will remove duplicate code.

symbolic-base/src/ZkFold/Base/Data/Conditional.hs
TurtlePU: OR we can ultimately decide that the sort of types which are `Conditional` on a certain boolean type `b` have to be decided by `b` itself, not by the types in sort, like
```haskell
class Boolean b => Conditional b where
  type Universe b :: Type -> Constraint
  ifThenElse :: Universe b a => b -> a -> a -> a
```
(name `Universe` is up to debate, of course, as this name is too generic)
Then we could have:
* `Universe Prelude.Bool` to be some universally satisfied constraint (already appeared in my [collection of thoughts on numerical hierarchy](https://gist.github.com/TurtlePU/55e0f3a8ff0929be84199bcf643de58e#file-poly-hs-L155), btw)
* `Universe (Bool ctx) a` to be `(SymbolicData a, Context a ~ ctx)` (this would require a trick with a phantom class too, though)

TurtlePU: Sorry that I didn't notice this earlier, only after witnessing this `Vector` instance x)

echatav: This same objection holds for `Conditional` instances derived for generic Haskell products, which will be less efficient than Haskell's `bool` or standard `if then else`, which is of course _still available_ for all types, Haskell or Symbolic. The change to `Conditional` you suggest is a major one.

TurtlePU: Oh, indeed...

vlasin: Here is my reasoning. For `x` such that `SymbolicData x`, there is only one choice of boolean: `Bool (Context x)`. So, ideally, there should be only one choice of `Eq` and `Conditional` for `x`.

Maybe we could override the generic implementation in the case of `Interpreter`? Would this solve the efficiency issue?

Regarding the elliptic curve hierarchy... The situation is different. As I noted, we have several options how the `BaseField` of the context can be related to the elliptic curve point type. Depending on the field choice, the implementation of the EC operations will be different. So, a multi-parametric approach is justified (and it is very much needed in the IVC, where we want to use all those options).

TurtlePU: > So, ideally, there should be only one choice of `Eq` and `Conditional` for `x`.

This would be of course ideal, but, for `Conditional`, this is not really optimal -- for `instance Conditional (Vector a)` we have the following dilemma:

* if we keep only one instance, then this should be an instance which builds new vector componentwise, as done in the PR. This is suboptimal in pure Haskell context; and we would still need instance for pure Haskell if we want to write down language-agnostic generalized algorithms, which is the point of @echatav's refactoring.

* if we need more instances, then we need a multiparam class or something of the sort.

Actually, I think the current multiparam class definition is fine, it's the instances that have to be changed to:

```haskell
class BoolType b => Conditional b a where
  ifThenElse :: b -> a -> a -> a

instance Conditional Haskell.Bool a where
  ifThenElse condition thenBranch elseBranch =
    if condition then thenBranch else elseBranch

instance SymbolicData a => Conditional (Bool (Context a)) a where
  ifThenElse (Bool c) t e = unPar1 $ interpolate [(zero, e), (one, t)] c
```

Where `interpolate` is a member of `SymbolicData` with the following type:
```haskell
class SymbolicData a where
  ...
  interpolate :: (Context a ~ c, Representable f) => [(BaseField c, a)] -> c f -> f a
```
(For now, `interpolate` can be defined only on lists of size 2 for simplicity of implementation)

The reasoning behind `interpolate` is such that once we have sum types in Symbolic, we will need to define an eliminator for them. For `Bool`, this is `Conditional`; to avoid defining a separate class for each sum type, I think it's better to include the common functionality in the base `SymbolicData` class and call into it.

TurtlePU: > return to the design of having `Conditional` and `Eq` being multiparameter typeclasses

Agreed about `Conditional` being multiparam, but I think `Eq` should still have an associated type, like this:
```
class Conditional (BooleanOf a) a => Eq a where
  type BooleanOf a :: Type
  (==) :: a -> a -> BooleanOf a
  (/=) :: a -> a -> BooleanOf a
```
My arguments are:
* Runtime performance doesn't differ much between `Haskell.Eq` and this `Eq`, even in pure context
* This makes _sense_ because while `Conditional` is like an eliminator for a boolean datatype, `Eq a` is a property of `a` that its equality relation is decidable (well, what "decidable" means in Symbolic is an interesting question, but still).
* On the topic of type inference. In the case of `Conditional`, having it multiparam is fine because all the types can be inferred from the inputs. Output of `Eq`, on the other hand, cannot be inferred from the inputs directly so we have to provide some help to inference algorithm in form of either fundeps or associated types.

TurtlePU:
Additional reasoning on why `BoolType`, `Conditional` and `Eq` have to be defined in different modules -- they mean completely different, although related concepts:
* `BoolType a` states that `a` is a boolean algebra
* `Conditional b a` states that `a` belongs to the universe where `b` can be eliminated like a boolean
* `Eq a` states that `a` has decidable equality relative to `BooleanOf a`

symbolic-base/src/ZkFold/Base/Data/Conditional.hs
echatav: ^ Yes, I agree that having its boolean be inferable from a type makes more sense for `Eq` than `Conditional`.

echatav:
Closing since we've basically finished with the logic hierarchy.

TurtlePU:
The branch can be deleted as well then, right?

PR #439, opened 2025-01-08
Added `SymbolicFold` and its instances

### LINKED ISSUE #431: Expose `fold` constraints via `MonadCircuit`/`Symbolic` API

TurtlePU:
Will resolve #431 once it's complete.

### DIFF
--- a/symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
+++ b/symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
@@ -48,9 +48,9 @@ algebraicMap Predicate {..} pi pm _ pad = padDecomposition pad f_sps_uni
         witness = M.fromList $ zip (keys $ acWitness predicateCircuit) (V.head pm)
 
         varMap :: SysVar i -> f
-        varMap (InVar inV)   = index pi inV
-        varMap (NewVar newV) = M.findWithDefault zero newV witness
-        varMap (FoldVar _ _) = P.error "FIXME"
+        varMap (InVar inV)            = index pi inV
+        varMap (NewVar (EqVar newV))  = M.findWithDefault zero newV witness
+        varMap (NewVar (FoldVar _ _)) = P.error "unexpected FOLD constraint"
 
         f_sps :: Vector (d+1) [PM.Poly a (SysVar i) Natural]
         f_sps = degreeDecomposition @d $ sys
--- a/symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
+++ b/symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
@@ -50,6 +50,7 @@ algebraicMap Predicate {..} pi pm _ pad = padDecomposition pad f_sps_uni
         varMap :: SysVar i -> f
         varMap (InVar inV)   = index pi inV
         varMap (NewVar newV) = M.findWithDefault zero newV witness
+        varMap (FoldVar _ _) = P.error "FIXME"
 
         f_sps :: Vector (d+1) [PM.Poly a (SysVar i) Natural]
         f_sps = degreeDecomposition @d $ sys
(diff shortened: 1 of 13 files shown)


symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
```diff
@@ -50,6 +50,7 @@ algebraicMap Predicate {..} pi pm _ pad = padDecomposition pad f_sps_uni
         varMap :: SysVar i -> f
         varMap (InVar inV)   = index pi inV
         varMap (NewVar newV) = M.findWithDefault zero newV witness
+        varMap (FoldVar _ _) = P.error "FIXME"
```
TurtlePU, line 53: @vlasin, what do we do here?
Is it expected that the input circuit has no fold constraints, maybe?

symbolic-base/src/ZkFold/Base/Protocol/IVC/SpecialSound.hs
```diff
@@ -57,7 +57,10 @@ specialSoundProtocol :: forall d a i p .
     ) => Predicate a i p -> SpecialSoundProtocol 1 i p [a] [a] a
 specialSoundProtocol phi@Predicate {..} =
   let
-      prover pi0 w _ _ = elems $ witnessGenerator predicateCircuit (pi0 :*: w) (predicateEval pi0 w)
+      prover pi0 w _ _ =
+        -- FIXME: what to do with values from folds?
+        let (wGen, _) = witnessGenerator predicateCircuit (pi0 :*: w) (predicateEval pi0 w)
```
TurtlePU, line 62: @vlasin, same question about fold constraints here

symbolic-base/src/ZkFold/Base/Protocol/IVC/AlgebraicMap.hs
vlasin: Yes, let's assume that. In fact, we currently ignore range constraints as well. Will need to fix it.

symbolic-base/src/ZkFold/Base/Protocol/IVC/SpecialSound.hs
vlasin: For now, let's assume that the IVC step circuit does not have those.

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -295,18 +341,35 @@ witToVar (WitnessF w) = runHash @(Just (Order a)) $ w $ \case
...
   WSysVar (NewVar newV) -> M newV
+  WSysVar (FoldVar fldID fldV) -> merkleHash (fldID, fldV)
 
 ----------------------------- Evaluation functions -----------------------------
 
 witnessGenerator ::
   (Arithmetic a, Representable p, Representable i) =>
-  ArithmeticCircuit a p i o -> p a -> i a -> Map ByteString a
+  ArithmeticCircuit a p i o -> p a -> i a ->
+  (Map ByteString a, Map ByteString (Map ByteString a))
```
vlasin, line 351: Could you explain why do we need to change the signature here?

TurtlePU: First map is for values of usual NewVar-s, second is for new FoldVar-s

TurtlePU: Maybe the cleaner approach would be to specify a new type of "created variables" (NewVar-s and FoldVar-s) and have a map from this type to the base field

vlasin: Yes, I agree. That would be preferable.

vlasin: However, is it possible to just use bytestrings for all types of variables?

TurtlePU: This would increase chances of collision between different types of variables so I'd prefer to use separate constructors

TurtlePU: I think a proper implementation of proof protocols would need to define their own circuit-like types and use whatever variable representation they need
(And, for example, the `ArithmeticCircuit.Map` module wouldn't be needed at all)

vlasin: Ok, let's leave it like this. Although, the collision problem can be solved by adding tags in the variable generation algorithm.

vlasin (approved):

hovanja2011 (approved):

Issue #426, opened 2025-01-06
Remove `HaskellCore` and `RustCore` from proving protocols


vlasin:
Refactor FFI-related code in zkfold-base according to @TurtlePU's suggestion. I.e.,remove HaskellCore and RustCore and prepare Plonk/Plonkup for the introduction of backend-specific field and curve types.

PR #440, opened 2025-01-13
feat: add fft poly div

diS3e:


vlasin:
What is the status of this PR? We discarded this approach in favor #426 ?

PR #441, opened 2025-01-14
Add Merkle tree to Symbolic

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -0,0 +1,186 @@
...
+  where
+    xP = leaf @c @x @d (V.last nodes) $ ind path
+
+    inits = V.unsafeToVector @(d-1) $ LL.unfoldr (\v -> if LL.null v then P.Nothing else P.Just (ind @_ @c (V.unsafeToVector v), LL.init v)) (V.fromVector path)
+
+    cinds :: Vector (d-1) (c Par1)
+    cinds = unpacked $ fromCircuit2F (pack path) (pack inits) $ \ps' is' -> do
+      let ps = P.fmap unPar1 (unComp1 ps')
+          is = V.unsafeToVector $ fromConstant @(BaseField c) zero : (P.init . V.fromVector $ P.fmap unPar1 (unComp1 is'))
+      mzipWithMRep (\wp wi -> newAssigned (one - ($ wp) + ($ wi)*(one + one))) ps is
```
TurtlePU, line 130: What does this formula mean?

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -0,0 +1,218 @@
...
+  deriving (Generic)
+
+hashAux :: forall i m a w f. (MonadCircuit i a w m, Representable f, Traversable f) => Par1 i -> f i -> f i -> m (f i)
+hashAux (Par1 b) h g = do
+  v1 <- merkleHasher h g
+  v2 <- merkleHasher g h
+  mzipWithMRep (\wx wy -> newAssigned ((one - ($ b)) * ($ wx) + ($ b) * ($ wy))) v1 v2
+  where
+    merkleHasher :: f i -> f i -> m (f i)
+    merkleHasher = P.undefined
```
TurtlePU, line 54: @vlasin, what is the expected function here?

TurtlePU (commented):
I think we can merge it after the hashing function is specified

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
vlasin: Ideally, it should be a parameter. But in this PR, we could use `mimcHash`.

vlasin: Ideally, it should be a parameter. But in this PR, we could use `mimcHash`.

vlasin: Ideally, it should be a parameter. But in this PR, we could use `mimcHash`.

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -0,0 +1,302 @@
...
+        _ -> let (n, ns) = L.uncons ls in n : helper @s ns (k-!1)
+
+hashAux :: forall c x.
+  ( Symbolic c
+  , SymbolicData x)
+  => c Par1 -> c (Layout x) -> c (Layout x) -> c (Layout x)
+hashAux b' h g = do
+  let (Vec v1) = merkleHasher [Vec h, Vec g]
+      (Vec v2) = merkleHasher [Vec g, Vec h]
+  fromCircuit3F b' v1 v2 $ \ (Par1 b) x y ->
```
TurtlePU, line 131: Here, an `ifThenElse` can be used from `instance Conditional (Bool c) (Vec f c)`

symbolic-base/src/ZkFold/Symbolic/Data/Vec.hs
```diff
@@ -0,0 +1,74 @@
...
+
+deriving newtype instance (Symbolic c, LayoutFunctor f) => Eq (Vec f c)
+deriving newtype instance (Symbolic c, LayoutFunctor f) => Conditional (Bool c) (Vec f c)
+
+instance (Symbolic c, Representable f) => FromConstant Natural (Vec f c) where
+    fromConstant = Vec . embed . tabulate . fromConstant
+instance (Symbolic c, Representable f) => FromConstant Integer (Vec f c) where
+    fromConstant = Vec . embed . tabulate . fromConstant
+
+instance {-# INCOHERENT #-} (Symbolic c) => FromConstant (Vec f c) (Vec f c)
```
TurtlePU, line 45: It's better to use `OVERLAPPING` instead of `INCOHERENT`

symbolic-base/src/ZkFold/Symbolic/Data/Vec.hs
```diff
@@ -0,0 +1,74 @@
...
+
+instance {-# INCOHERENT #-} (Symbolic c, Representable f, a ~ BaseField c) => FromConstant a (Vec f c) where
```
TurtlePU, line 47: It's better to use `OVERLAPPING` instead of `INCOHERENT`

TurtlePU:
From the first look at the PR, seems that the `foldr` test for lists which failed before was fixed in #550, `lookup` test highlighted an issue with `Vector -> MerkleTree` conversion implementation, and `find` test takes too much time both locally and in CI, investigating

TurtlePU:
(Also will work on resolving of some of my comments from before :D)

vlasin (approved):

vlasin (approved):

PR #446, opened 2025-01-16
Add BooleanOf type family to Eq

echatav:
This PR re-introduces the `BooleanOf` type family that had previously been a part of the `EllipticCurve` class, and puts it into a more natural setting in the `Eq` class.


TurtlePU (approved):
:shipit:

PR #456, opened 2025-01-21
CycleFold prototype implementation

vlasin:
I drafted a new update to IVC that implements the CycleFold algorithm. I had to mock the elliptic curve cycle and foreign arithmetic with `Par1` as we are waiting for the respective PRs.


TurtlePU:
Sorry for the diff here, maybe you might want to roll back these commits (anyway the development is preserved in #488)

PR #458, opened 2025-01-24
Update tests

vlasin:
It's not a full test refactor, but I group tests in a more logical way.


symbolic-base/symbolic-base.cabal
```diff
@@ -252,36 +252,36 @@ test-suite symbolic-base-test
...
-      Tests.Permutations
-      Tests.Plonkup
-      Tests.ReedSolomon
-      Tests.RSA
-      Tests.SHA2
-      Tests.UInt
-      Tests.Univariate
-      Tests.Univariate.Poly
-      Tests.Univariate.PolyVec
+      Tests.Base.Algebra.EllipticCurve
```
TurtlePU, line 255: Maybe we could start removing `Base` and `Basic` prefixes from here? For example, this module would be named `Tests.Algebra.EllipticCurve`

symbolic-base/test/Main.hs
```diff
@@ -1,46 +1,52 @@
...
+import           Tests.Symbolic.ArithmeticCircuit        (specArithmeticCircuit)
+import           Tests.Symbolic.Compiler                 (specCompiler)
+import           Tests.Symbolic.Data.ByteString          (specByteString)
+import           Tests.Symbolic.Data.FFA                 (specFFA)
+import           Tests.Symbolic.Data.Hash                (specHash)
+import           Tests.Symbolic.Data.List                (specList)
+import           Tests.Symbolic.Data.UInt                (specUInt)
 
 main :: IO ()
 main = do
```
TurtlePU, line 29: I guess it would also be more idiomatic to refactor `Tests.*` modules so that they would export [Spec](https://hackage.haskell.org/package/hspec-2.11.10/docs/Test-Hspec.html#t:Spec)s instead of `IO ()` actions and run them inside a single toplevel `hspec`.

echatav (dismissed):
LGTM

symbolic-base/symbolic-base.cabal
vlasin: Agree, thanks!

TurtlePU (approved):

PR #467, opened 2025-01-29
New `divMod` algorithm for UInt

TurtlePU:
Part of FFA refactor


symbolic-base/src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -475,7 +475,8 @@ instance AdditiveMonoid Natural where
 instance Semiring Natural
 
 instance SemiEuclidean Natural where
-    divMod = Haskell.divMod
+    divMod m 0 = (0, m)
```
vks4git, line 478: Is this specific behaviour required somewhere?

symbolic-examples/stats/UInt.DivMod.32.Auto
```diff
@@ -1,4 +1,4 @@
 UInt.DivMod.32.Auto
-Number of constraints: 999
-Number of variables: 1347
-Number of range lookups: 766
\ No newline at end of file
+Number of constraints: 96
```
vks4git, line 2: Awesome!

symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -265,33 +273,48 @@ instance
...
-                -> (UInt n rs c, UInt n rs c)
-            longDivisionStep (q', r') i =
-                let rs = force $ addBit (r' + r') (value @n -! i -! 1)
-                 in bool @(Bool c) (q', rs) (q' + fromConstant ((2 :: Natural) ^ i), rs - d) (rs >= d)
+instance ( Symbolic c, KnownNat n, KnownRegisterSize r
+         , KnownRegisters c n r
+         , regSize ~ GetRegisterSize (BaseField c) n r
+         , KnownNat (Ceil regSize OrdWord)
+         ) => SemiEuclidean (UInt n r c) where
+    divMod num@(UInt nm) den@(UInt dn) =
```
vks4git, line 281: Can you please add a few comments about what each function inside `where` clause does and how everything works in general? I get the general idea but not the details

symbolic-base/src/ZkFold/Base/Algebra/Basic/Class.hs
TurtlePU: Yeah, division by zero happens when a test for Symbolic `divMod` runs
Now that I think about it, maybe it is better to modify the test so that it doesn't choose zero as a denominator

symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
TurtlePU: Added!

symbolic-examples/stats/UInt.DivMod.32.Auto
zlonast, line 2: Does this only affect this benchmark because the others don't use division?

TurtlePU: Seems like it, yes
Benchmarks are run in CI so code checks wouldn't pass if anything else changed

TurtlePU: We have some benchmarks commented out though because they used too much RAM to compile, maybe some of them started working now

vks4git (approved):

vlasin (approved):

PR #470, opened 2025-01-31
Vks4git/google signature

vks4git:


symbolic-examples/src/ZkFold/Symbolic/Examples.hs
```diff
@@ -84,6 +84,7 @@ examples =
   , ("Fibonacci.100", exampleOutput $ exampleFibonacci 100)
   , ("MiMCHash", exampleOutput exampleMiMC)
   , ("SHA256.32", exampleOutput $ exampleSHA @32)
+  -- , ("JWT.secretBits", exampleOutput $ exampleJWTSerialisation)
```
vlasin, line 87: Does this bench work?

symbolic-cardano/src/ZkFold/Symbolic/Cardano/Contracts/ZkLogin.hs
```diff
@@ -1,54 +1,35 @@
...
+    .  RSA ctx 10328
+    => SecretBits ctx
+    => ClientSecret ctx
+    -> ByteString 64 ctx
+    -> ByteString 256 ctx
+    -> Certificate ctx
+    -> PublicInput ctx
     -> Bool ctx
-zkLogin Certificate{..} token@UserToken{..} date clientId = sigValid && dateValid && audValid
+zkLogin clientSecret@ClientSecret{..} amount recipient certificate pi = tokenValid && piValid
```
vlasin, line 29: I am not sure why we need the last argument for this function. We can derive it from the other arguments.

symbolic-examples/src/ZkFold/Symbolic/Examples.hs
vks4git: It does locally, but it kills CI

symbolic-base/src/ZkFold/Symbolic/Data/VarByteString.hs
```diff
@@ -0,0 +1,356 @@
...
+import           ZkFold.Symbolic.Data.Input        (SymbolicInput)
+import           ZkFold.Symbolic.Interpreter
+import           ZkFold.Symbolic.MonadCircuit      (MonadCircuit, newAssigned)
+
+-- | A ByteString that has length unknown at compile time but guaranteed to not exceed @maxLen@.
+-- The unassigned buffer space (i.e. bits past @bsLength@) should be set to zero at all times.
+--
+-- TODO: Declare all the instances ByteString has for VarByteString
+--
+data VarByteString (maxLen :: Natural) (context :: (Type -> Type) -> Type) =
```
vlasin, line 61: Maybe I missed the discussion. Why do we need this type in addition to a fixed length bytestring? Is it more efficient in some way?

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
```diff
@@ -0,0 +1,457 @@
...
+import           ZkFold.Symbolic.Data.Combinators
+import           ZkFold.Symbolic.Data.Eq
+import           ZkFold.Symbolic.Data.FieldElement
+import           ZkFold.Symbolic.Data.Input         (SymbolicInput)
+import           ZkFold.Symbolic.Data.Ord
+import           ZkFold.Symbolic.Data.UInt
+import qualified ZkFold.Symbolic.Data.VarByteString as VB
+import           ZkFold.Symbolic.Data.VarByteString (VarByteString (..), wipeUnassigned, (@+))
+import           ZkFold.Symbolic.MonadCircuit       (newAssigned)
+
```
vlasin, line 53: Should we add some tests for this module? Perhaps a signature verification test?

symbolic-base/src/ZkFold/Symbolic/Data/VarByteString.hs
vks4git: In a token, there are several fields that have variable length (name, email, issuer, audience, etc.). This type provides an interface to work with such fields, for example, concatenating two VarByteStrings will move all the unassigned space to the left. 
Also, yes, I plan to make the underlying container a vector of words instead of bits to optimise shifts

vlasin: It seems to me, instead of introducing a new bytestring type, we just need a function that constructs a token from data. I.e., only the concatenation operation is actually required. Is it correct?

vks4git: If we just construct a token from data, what will its true length be? It should also be a VarByteString, right. Also, we need to calculate SHA256 of such strings (the necessary modifications were introduced). 
The motivation of making VarByteString a separate module was that it might be useful to prospective users of Symbolic if they need a way to work with bytestrings of unknown length. Moreover, it just provides a convenient way to store a bytestring and its true length.

vlasin: Ok. I guess, in the future, we could use folding to reduce the number of constraints and make it more convenient to work with.

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
vks4git: Done

vlasin (approved):

vlasin (approved):

PR #478, opened 2025-02-05
Replace x with gs

zlonast:


symbolic-base/src/ZkFold/Base/Protocol/Plonkup/Internal.hs
```diff
@@ -41,15 +43,16 @@ with4n6 f = withDict (timesNat @4 @n) (withDict (plusNat @(4 * n) @6) f)
...
-  ( KnownNat n, Arithmetic (ScalarFieldOf g1), Arbitrary (ScalarFieldOf g1)
+  ( KnownNat n, KnownNat (n + 5), Arbitrary g1, Arbitrary g2, Arithmetic (ScalarFieldOf g1), Arbitrary (ScalarFieldOf g1)
   , Arbitrary (ArithmeticCircuit (ScalarFieldOf g1) p i l)
   ) => Arbitrary (Plonkup p i n l g1 g2 t) where
     arbitrary = do
         ac <- arbitrary
+        h1 <- arbitrary
         let (omega, k1, k2) = getParams (value @n)
-        Plonkup omega k1 k2 ac <$> arbitrary
+        Plonkup omega k1 k2 ac h1 <$> arbitrary
```
vlasin, line 58: Is it really correct? I think here we should first create `x`, then set `h1` and `gs`.

vlasin (approved):

PR #481, opened 2025-02-07
Ord hierarchy reorganization

### LINKED ISSUE #23: Update `Ord` class definition
See module `ZkFold.Symbolic.Data.Ord` for details.

echatav:
Addresses #420.

* Allows generic derivation of `Ord`
* Reduces the number of constraints for field element inequality!
* Also resolves #23

For compatibility, `UInt`s `Ord` instance implementation was left unchanged; see #486 for followup.


symbolic-base/src/ZkFold/Symbolic/Data/Ord2.hs
```diff
@@ -0,0 +1,200 @@
...
+bitwiseCompare :: forall c . Symbolic c => c [] -> c [] -> Ordering c
+bitwiseCompare x y = fold ((zipWith (compare `on` Bool) `on` unpacked) x y)
+
+getBitsBE :: forall c f . (Symbolic c, LayoutFunctor f) => c f -> c []
+-- ^ @getBitsBE x@ returns a list of circuits computing bits of @x@, eldest to
+-- youngest.
+getBitsBE x = symbolicF (arithmetize x Proxy)
+    (concatMap (reverse . padBits n . map fromConstant . binaryExpansion . toConstant))
+    (fmap (concatMap reverse) . traverse (expansion n) . toList)
+  where n = numberOfBits @(BaseField c)
```
echatav, line 160: Can this be simplified?

echatav: It doesn't need to use `[]`, any `Foldable` container of bits should be fine

echatav: Like maybe it's in `f :*: Vector (NumberOfBits (BaseField c)` or something?

vlasin: I think, it might be a thing!

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
```diff
@@ -1,158 +1,201 @@
...
+bitwiseCompare :: forall c . Symbolic c => c [] -> c [] -> Ordering c
+bitwiseCompare x y = fold ((zipWith (compare `on` Bool) `on` unpacked) x y)
+
+getBitsBE :: forall c f . (Symbolic c, LayoutFunctor f) => c f -> c []
 -- ^ @getBitsBE x@ returns a list of circuits computing bits of @x@, eldest to
 -- youngest.
 getBitsBE x = symbolicF (arithmetize x Proxy)
     (concatMap (reverse . padBits n . map fromConstant . binaryExpansion . toConstant))
     (fmap (concatMap reverse) . traverse (expansion n) . toList)
   where n = numberOfBits @(BaseField c)
```
echatav, line 161: Perhaps these should look like:
```Haskell
bitwiseCompare :: forall c f . (Symbolic c, Foldable f, Representable f) => f (Bool c) -> f (Bool c) -> Ordering c

getBitsLayout :: SymbolicOutput x => x
  -> (Layout x :.: Vector (NumberOfBits (BaseField (Context x)))) (Bool (Context x))
```

echatav: ^ maybe in a followup PR?

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
```diff
@@ -1,158 +1,201 @@
...
+  ordering x y z o = bool (bool x y (o == eq)) z (o == gt)
+  compare = bitwiseCompare `on` getBitsBE
+
+bitwiseCompare :: forall c . Symbolic c => c [] -> c [] -> Ordering c
+bitwiseCompare x y = fold ((zipWith (compare `on` Bool) `on` unpacked) x y)
+
+getBitsBE :: forall c f . (Symbolic c, LayoutFunctor f) => c f -> c []
 -- ^ @getBitsBE x@ returns a list of circuits computing bits of @x@, eldest to
 -- youngest.
 getBitsBE x = symbolicF (arithmetize x Proxy)
```
TurtlePU, line 158: We don't need to `arithmetize x` because `x` is already a `c f`

symbolic-base/src/ZkFold/Symbolic/Data/Ord2.hs
TurtlePU: > Is this a thing?

It can be done, but I'm afraid this would be very ineffective as it would
1. Lump all the constraints from fields of `x` together in a single `Context x (Layout x)` circuit
2. Split them apart in a `Layout x :.: Vector ...` so that each bit would then contain constraints from the whole circuit

So at least `Context x (Layout x :.: Vector _)` would be better, though I don't know if we need it here

TurtlePU (dismissed):

symbolic-base/src/ZkFold/Symbolic/Data/Ord.hs
echatav: removed `arithmetize`

TurtlePU (approved):

PR #488, opened 2025-02-17
Changes to NativeContext

TurtlePU:
A concept on how to make cyclefold work
Doesn't compile!
Key idea: we need two Symbolic point types from curve cycle where all fields are FFA.
Thanks to how our FFA works, in case its order matches the base field of a circuit, it is represented with FieldElement internally so is very cheap.
So by choosing appropriate contexts for these point types we obtain datatypes needed for CycleFold.
To compute layout, you can use `Interpreter a` context.
To compute FFA without storing constraints, we might need a `FFAW` newtype similar to `FieldElementW` (or wait for Symbolic refactor and play with contexts).
To convert values, you can use `ToConstant (FFA p r (Interpreter a))` instance.

### DIFF
--- a/symbolic-base/src/ZkFold/Base/Data/Empty.hs
+++ b/symbolic-base/src/ZkFold/Base/Data/Empty.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE TypeOperators #-}
+
+module ZkFold.Base.Data.Empty where
+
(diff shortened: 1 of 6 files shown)


TurtlePU:
Isn't needed since a solution with another approach was already merged

Issue #486, opened 2025-02-08
Simplify Haskell constraints for `instance Ord (UInt n r c)`


TurtlePU:
Currently, `instance Ord (UInt n r c)` works by splitting the registers of size `r` into registers of size `OrdWord = 16`. I think that actually we want `OrdWord` to be equal to the main register size of UInt because
* most likely, `r` would be set to 16 anyway
* there's no natural way for the user to regulate this `OrdWord` parameter except for maybe introducing yet another register parameter for `UInt` type
* an optimized version of `UInt` would need to use lookup arguments for both its registers, so in case they're different the size of the witness table is `2 ^ r + 2 ^ OrdWord` instead of just `2 ^ r`.

So I propose to get rid of this split and use usual registers of size `r` inside `Ord`. As a bonus, this would once again simplify the type-level constraints needed to work with `UInt`.

* [x] add benchmark for `UInt` `>`
* [x] add correctness test for `UInt` `>`

vlasin:
Yes!

vlasin:
I think we came to an agreement that the use of `Auto` registers + small virtual registers for specific operations is better. Can we close this one?
@TurtlePU @vks4git

TurtlePU:
The issue looks outdated anyway, so yeah, let's close it

PR #495, opened 2025-02-27
UInt Ord

echatav:
See #486

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
@@ -39,7 +39,7 @@ import           Data.Map                          (fromList, (!))
 import           Data.Traversable                  (for, traverse)
 import           Data.Tuple                        (swap)
 import qualified Data.Zip                          as Z
-import           GHC.Generics                      (Generic, Par1 (..), (:*:) (..))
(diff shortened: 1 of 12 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -412,6 +413,18 @@ instance ( Symbolic c, KnownNat n, KnownRegisterSize r
...
+    .  Symbolic ctx
+    => KnownNat regSize
+    => ctx (Vector k)                           -- @k@ registers of size up to @regSize@
+    -> ctx (Vector k :.: Vector regSize) -- @k * wordsPerReg@ registers of size @wordSize@
+asWordsReg v = fromCircuitF v $ \regs -> do
+    words <- for regs $ \reg -> do
+      bits <- expansionW @regSize (value @regSize) reg
+      return (V.reverse (V.unsafeToVector bits))
+    return (Comp1 words)
+
```
echatav, line 427: Not sure if this is the right way to expand. Is there a better way? I'm concerned about it incurring proliferating `KnownNat` constraints which is nothing new but annoying and feels like it shouldn't be necessary.

echatav:
Closing in favor of #512

PR #503, opened 2025-03-03
UPLC Converter M3

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+import           ZkFold.Symbolic.Data.Class       (SymbolicData)
+import           ZkFold.Symbolic.Data.Combinators
+import           ZkFold.Symbolic.Data.Conditional
+import           ZkFold.Symbolic.Data.Eq
+import           ZkFold.Symbolic.Data.Ord
+import           ZkFold.Symbolic.Data.UInt
+import           ZkFold.Symbolic.Interpreter      (Interpreter (..))
+
+
+newtype Int (n :: Natural) (r :: RegisterSize) (c :: (Type -> Type) -> Type) = Int { uint :: UInt (n+1) r c}
```
TurtlePU, line 32: I think it's better to use `UInt n r c` to reuse same type-level constraints from `UInt` and to make the API more intuitive. Compare it with usual `uint32_t`/`int32_t` from C, for example.

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+
+deriving instance Generic (Int n r c)
+deriving instance (NFData (c (Vector (NumberOfRegisters (BaseField c) (n+1) r)))) => NFData (Int n r c)
+deriving instance (Haskell.Eq (c (Vector (NumberOfRegisters (BaseField c) (n+1) r)))) => Haskell.Eq (Int n r c)
+deriving instance (Haskell.Show (BaseField c), Haskell.Show (c (Vector (NumberOfRegisters (BaseField c) (n+1) r))))
```
TurtlePU, line 37: Why is `Haskell.Show (BaseField c)` here? Is it required?

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+    => Haskell.Show (Int n r c)
+deriving instance (KnownRegisters c (n+1) r, Symbolic c) => SymbolicData (Int n r c)
+deriving instance (KnownRegisters c (n+1) r, Symbolic c) => Conditional (Bool c) (Int n r c)
+deriving instance (KnownRegisters c (n+1) r, Symbolic c) => Eq (Int n r c)
+
+instance (Symbolic c, KnownNat n, KnownRegisterSize r) => FromConstant Natural (Int n r c) where
+    fromConstant c = withDict (plusNat @n @1) $ Int (fromConstant c)
+
+instance (Symbolic c, KnownNat n, KnownRegisterSize r) => FromConstant Integer (Int n r c) where
+    fromConstant c = withDict (plusNat @n @1) $ Int (Haskell.bool (fromConstant c) (negate $ fromConstant (Haskell.negate c)) (c Haskell.<0))
```
TurtlePU, line 47: `negate (fromConstant (negate c)) = fromConstant c` by laws of `fromConstant`, and instance for `UInt` is lawful

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+
+instance MultiplicativeMonoid (Int n r c) => Exponent (Int n r c) Natural where
+    (^) = natPow
+
+------------------------------------------------------------------------------
+
+instance
+    ( Symbolic (Interpreter a), KnownNat n, KnownRegisterSize r
```
TurtlePU, line 55: `Symbolic (Interpreter a)` holds if `Arithmetic a`, it's better to use it here

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+
+isNegative :: forall n r c.
+    ( Symbolic c, KnownNat n, KnownRegisterSize r
+    ) => Int n r c -> Bool c
+isNegative (Int u) =
+    withDict (plusNat @n @1) $
+        withNumberOfRegisters @(n+1) @r @(BaseField c) $
+            withGetRegisterSize @(n+1) @r @(BaseField c) $
+                withCeilRegSize @(GetRegisterSize (BaseField c) (n+1) r) @OrdWord $
+                    u >= fromConstant ((2 :: Natural) ^ (value @n))
```
TurtlePU, line 72: This should optimize away to taking the highest bit in the representation. It would be clearer if we explicitly did this here.

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+        (Int u1, Int u2) = (abs i1, abs i2)
+        (d, m) = withGetRegisterSize @(n+1) @r @(BaseField c) $
+                    withCeilRegSize @(GetRegisterSize (BaseField c) (n+1) r) @OrdWord $
+                        divMod u1 u2
+
+
+        ite_tf = ifThenElse (isNegative i1 && not (isNegative i2)) dm_mp ite_pm
+        ite_pm = ifThenElse (not (isNegative i1) && isNegative i2) dm_pm dm_pp
+
+        dm_mm = (Int d, negate (Int m))
```
TurtlePU, line 129: Modulus is always positive by definition, I think we should stick to the definition

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+
+        ite_tf = ifThenElse (isNegative i1 && isNotNegative i2) dm_mp ite_pm
+        ite_pm = ifThenElse (isNotNegative i1 && isNegative i2) dm_pm dm_pp
+
+        dm_mm = (Int d, negate (Int m))
+        dm_mp = (negate (Int d), negate (Int m))
+        dm_pm = (negate (Int d), Int m)
+        dm_pp = (Int d, Int m)
+
+qout :: forall n r c.
```
TurtlePU, line 167: typo

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,216 @@
...
+    i1@(Int u1) >= i2@(Int u2) = pm || (ub && (pp || nn))
+        where
+            ub = withGetRegisterSize @(n+1) @r @(BaseField c) $
+                    withCeilRegSize @(GetRegisterSize (BaseField c) (n+1) r) @OrdWord $
+                        u1 >= u2
+            nn = isNegative i1 && isNegative i2
+            pp = isNotNegative i1 && isNotNegative i2
+            pm = isNotNegative i1 && isNegative i2
+
+    i1@(Int u1) > i2@(Int u2) = pm || (ub && (pp || nn))
```
TurtlePU, line 204: You can use `xor` here to simplify logic a bit

symbolic-base/test/Tests/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,187 @@
...
+    .  PrimeField (Zp p)
+    => KnownNat n
+    => KnownNat (n+1)
+    => KnownRegisterSize rs
+    => r ~ NumberOfRegisters (Zp p) (n+1) rs
+    => KnownNat r
+    => Spec
+specInt' = do
+    let n = value @n
+        m = 2 ^ (11 :: Natural)
```
TurtlePU, line 99: Why not `2 ^ n`?

symbolic-base/test/Tests/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,187 @@
...
+            x <- tossp m
+            return $ evalBoolVec @(Zp p) (isNegative @n @rs $ fromConstant x) === 0 .&.
+                     evalBoolVec @(Zp p) (isNegative @n @rs $ fromConstant (- x)) === 1
+        it "IsNegative correct-2" $ do
+            x <- toss m
+            let t = evalBoolVec @(Zp p) (isNegative @n @rs $ fromConstant x)
+            return $ (t === 0) .||. (t === 1)
+        it "Abs correct" $ do
+            x <- tossp m
+            return $ toConstant (abs (fromConstant (-x) :: Int n rs (Interpreter (Zp p)))) === x
```
TurtlePU, line 116: I think for a sanity check we can also check here that `abs x == x` and `abs 0 == 0`

symbolic-base/test/Tests/Symbolic/Data/Int.hs
```diff
@@ -0,0 +1,187 @@
...
+            num <- toss m
+            d <- toss1 m
+            let (acQ, acR) = (fromConstant num :: Int n rs (AC (Zp p))) `divMod` fromConstant d
+                (zpQ, zpR) = (fromConstant num :: Int n rs (Interpreter (Zp p))) `divMod` fromConstant d
+                (trueQ, trueR) = num `divMod` d
+                refQ = execZpInt (fromConstant trueQ ::Int n rs (Interpreter (Zp p)))
+                refR = execZpInt (fromConstant trueR ::Int n rs (Interpreter (Zp p)))
+            return $ (execAcInt acQ, execAcInt acR) === (execZpInt zpQ, execZpInt zpR) .&. (refQ, refR) === (execZpInt zpQ, execZpInt zpR)
+
+        when (m > 0) $ it "performs quotReq correctly" $ do
```
TurtlePU, line 143: typo

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -48,6 +66,7 @@ import           ZkFold.UPLC.Term
 -- and its interpretation as a Symbolic datatype in arbitrary context 'c'.
 class
     ( Typeable v, SymbolicData v
+    , SymbolicOutput v, SymbolicFold c
```
TurtlePU, line 69: `SymbolicOutput` implies `SymbolicData`

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,6 +298,28 @@ applyPoly ctx SndPair [arg] = do
...
+  return $ MaybeValue (symMaybe p (L.head v))
+applyPoly ctx (BPFList TailList) (xs0:_) = do
+  MaybeValue p <- evalArg ctx xs0 []
+  ExList v <- asList (Symbolic.fromJust p)
+  return $ MaybeValue (symMaybe p (L.tail v))
+applyPoly ctx (BPFList NullList) (xs0:_) = do
+  MaybeValue p <- evalArg ctx xs0 []
+  ExList v <- asList (Symbolic.fromJust p)
+  return $ MaybeValue (symMaybe p (L.null v))
 applyPoly _ (BPFList _) _ = error "FIXME: UPLC List support"
```
TurtlePU, line 323: This case can be removed? Or are there other list functions which are not supported yet?

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -319,11 +394,153 @@ instance
...
+evalMono (BMFInteger SubtractInteger)             = subtractIntegerFun @c
+evalMono (BMFInteger MultiplyInteger)             = multiplyIntegerFun @c
+evalMono (BMFInteger DivideInteger)               = divideIntegerFun @c
+evalMono (BMFInteger ModInteger)                  = modIntegerFun @c
+evalMono (BMFInteger QuotientInteger)             = quotientIntegerFun @c
+evalMono (BMFInteger RemainderInteger)            = remainderIntegerFun @c
+evalMono (BMFInteger EqualsInteger)               = equalsIntegerFun @c
+evalMono (BMFInteger LessThanInteger)             = lessThanIntegerFun @c
+evalMono (BMFInteger LessThanEqualsInteger)       = lessThanEqualsIntegerFun @c
+evalMono (BMFInteger _)                           = error "FIXME: UPLC Integer support"
```
TurtlePU, line 410: Are there any other `Int` functions which are not supported yet?

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -319,11 +394,153 @@ instance
...
-evalMono (BMFByteString _) = error "FIXME: UPLC ByteString support"
-evalMono (BMFString _)     = error "FIXME: UPLC String support"
-evalMono (BMFAlgorithm _)  = error "FIXME: UPLC Algorithms support"
-evalMono (BMFData _)       = error "FIXME: UPLC Data support"
-evalMono (BMFCurve _)      = error "FIXME: UPLC Curve support"
-evalMono (BMFBitwise _)    = error "FIXME: UPLC ByteString support"
+evalMono :: forall c s t.
+  ( Sym c
+  ) => BuiltinMonoFunction s t -> Fun s t c
+evalMono (BMFInteger AddInteger)                  = addIntegerFun @c
```
TurtlePU, line 400: What does the compiler error say when you remove the `@c` application?

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,6 +298,28 @@ applyPoly ctx SndPair [arg] = do
...
+applyPoly ctx (BPFList ChooseList) (ct:tt:et:args) = let
+    b0 = applyPoly ctx (BPFList NullList) [ct]
+    bt =   evalArg ctx (AThunk . Right $ b0) []
+  in applyPoly ctx IfThenElse ((AThunk . Right $ bt):tt:et:args)
+applyPoly ctx (BPFList MkCons) (x0:xs0:_) = do
+  MaybeValue x <- evalArg ctx x0 []
+  let l = Symbolic.fromJust x
+  MaybeValue xs <- evalArg ctx xs0 []
+  let ls = Symbolic.fromJust xs
+  return $ MaybeValue (Symbolic.just $ l L..: (fromJust $ cast ls))
```
TurtlePU, line 310: What about error handling? What if either of `x` or `xs` is a `Symbolic.nothing`?

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,6 +298,28 @@ applyPoly ctx SndPair [arg] = do
   MaybeValue p <- evalArg ctx arg []
   (_, ExValue v) <- asPair (Symbolic.fromJust p)
   return $ MaybeValue (symMaybe p v)
+applyPoly ctx (BPFList ChooseList) (ct:tt:et:args) = let
+    b0 = applyPoly ctx (BPFList NullList) [ct]
+    bt =   evalArg ctx (AThunk . Right $ b0) []
+  in applyPoly ctx IfThenElse ((AThunk . Right $ bt):tt:et:args)
+applyPoly ctx (BPFList MkCons) (x0:xs0:_) = do
```
TurtlePU, line 305: I think we expect no more arguments after `xs0`, so we should raise an error if there are more arguments, not ignore them

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,6 +298,28 @@ applyPoly ctx SndPair [arg] = do
...
+  MaybeValue x <- evalArg ctx x0 []
+  let l = Symbolic.fromJust x
+  MaybeValue xs <- evalArg ctx xs0 []
+  let ls = Symbolic.fromJust xs
+  return $ MaybeValue (Symbolic.just $ l L..: (fromJust $ cast ls))
+applyPoly ctx (BPFList HeadList) (xs0:_) = do
```
TurtlePU, line 311: `[xs0]`

TurtlePU, line 311: The rest of the arguments have to be applied to the result of `head` and shouldn't be ignored

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -293,6 +362,12 @@ evalConstant (CG2 _)         = error "FIXME: UPLC BLS support"
 pair :: Sym c => SymValue t c -> SymValue u c -> SymValue (BTPair t u) c
 pair (SymValue p) (SymValue q) = SymValue (p, q)
 
+makeSingleton :: (Sym c) => SymValue u c -> SymValue (BTList u) c
+makeSingleton (SymValue xs) = SymValue $ L.singleton xs
+
+concatList :: (Sym c) => SymValue u c -> SymValue (BTList u) c -> SymValue (BTList u) c
```
TurtlePU, line 368: Maybe `consList` would be a bit better name, `concat` usually means concatenation of two lists

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -277,14 +344,16 @@ data SymValue t c = forall v. IsData t v c => SymValue v
...
-evalConstant (CInteger _)    = error "FIXME: UPLC Integer support"
-evalConstant (CByteString _) = error "FIXME: UPLC ByteString support"
-evalConstant (CString _)     = error "FIXME: UPLC String support"
+evalConstant (CInteger i)    = withNumberOfRegisters @IntLength' @Auto @(BaseField c) $ SymValue (fromConstant i)
+evalConstant (CByteString b) = SymValue (fromConstant b)
+evalConstant (CString s)     = SymValue (fromString $ unpack s)
 evalConstant (CUnit ())      = SymValue Proxy
 evalConstant (CData _)       = error "FIXME: UPLC Data support"
-evalConstant (CList _)       = error "FIXME: UPLC List support"
+evalConstant (CList [])      = error "FIXME: UPLC Data support"
```
TurtlePU, line 354: `List` support, not `Data` support

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -56,13 +75,16 @@ class
     , Representable (Payload v)
     ) => IsData (t :: BuiltinType) v c | t c -> v, v -> t, v -> c where
   asPair :: v -> Maybe (ExValue c, ExValue c)
+  asList :: v -> Maybe (ExList c)
+
+data ExList c = forall t v. IsData t v c => ExList (L.List c v)
```
TurtlePU, line 80: A comment wouldn't hurt

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -204,17 +226,40 @@ applyMono b (FLam f) (v:args) = do
 -- This part is meant to be changed when completing the Converter implementation.
 
 -- Uncomment these lines as more types are available in Converter:
```
TurtlePU, line 228: This comment can be removed, I guess

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
hovanja2011: Yes, there are two functions (IntegerToByteString, ByteStringToInteger) that are not done (and in this task they do not need to be done)

symbolic-base/src/ZkFold/Symbolic/Data/Int.hs
hovanja2011: This is required for the corresponding UInt instance.

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
TurtlePU: Around _lists of_ IsData Symbolic types, I guess

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,7 +299,28 @@ applyPoly ctx SndPair [arg] = do
...
+applyPoly ctx (BPFList ChooseList) (ct:tt:et:args) = let
+    b0 = applyPoly ctx (BPFList NullList) [ct]
+    bt =   evalArg ctx (AThunk . Right $ b0) []
+  in applyPoly ctx IfThenElse ((AThunk . Right $ bt):tt:et:args)
+applyPoly ctx (BPFList MkCons) [x0,xs0] = do
+  MaybeValue x <- evalArg ctx x0 []
+  let l = Symbolic.fromJust x
+  MaybeValue xs <- evalArg ctx xs0 []
+  let ls = Symbolic.fromJust xs
+  return $ MaybeValue (symMaybe2 x xs $ l L..: (fromJust $ cast ls))
```
TurtlePU, line 311: Instead of `fromJust`, I think we should run cast inside the Maybe monad here:
```haskell
ls' <- cast ls
... symMaybe2 x xs $ l L..: ls' ...
```

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -253,7 +299,28 @@ applyPoly ctx SndPair [arg] = do
...
+applyPoly ctx (BPFList HeadList) (xs0:args) = do
+  MaybeValue p <- evalArg ctx xs0 args
```
TurtlePU, line 313: No, `args` should be applied to the head, not to the whole list
For now it would be better to just say that we don't want any more arguments with `[xs0]` in the pattern match then, I guess

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -319,11 +397,144 @@ instance
...
+
+sliceByteStringFun :: forall c .(Sym c) => Fun [BTInteger, BTInteger, BTByteString] BTByteString c
+sliceByteStringFun = withNumberOfRegisters @IntLength @Auto @(BaseField c) $ withIsoFieldElemUInt @IntLength @c $
+  fromConstant (\(Int f :: Int IntLength Auto c) (Int t :: Int IntLength Auto c) (VarByteString l b)
+                    -> let (fr, to) = (Symbolic.max (from f) zero, Symbolic.max (from f + from t-one) (l - one))
+                           fe8 = fromConstant (8::Natural) :: FieldElement c
+                           newB = shiftR (shiftL b (fromConstant (value @BSLength) - l + fr*fe8)) (fromConstant (value @BSLength) - to*fe8)
+                        in fromConstant $ Symbolic.just @c (VarByteString to newB) :: (Fun '[] BTByteString c))
+
+isoFieldElemUInt :: forall n c. (Symbolic c) => Dict (NumberOfBits (BaseField c) ~ n)
```
TurtlePU, line 495: This is definitely not true

symbolic-uplc/src/ZkFold/Symbolic/UPLC/Evaluation.hs
```diff
@@ -319,11 +397,144 @@ instance
...
+
+lengthOfByteStringFun :: forall c. (Sym c) => Fun '[BTByteString] BTInteger c
+lengthOfByteStringFun = withNumberOfRegisters @IntLength @Auto @(BaseField c) $ withIsoFieldElemUInt @IntLength @c $
+  fromConstant (\(VarByteString l _) -> Symbolic.just @c $ div (from (from l :: UInt IntLength Auto c)) (fromConstant (8 :: Natural)))
+
+indexByteStringFun :: forall c .(Sym c) => Fun [BTByteString, BTInteger] BTInteger c
+indexByteStringFun = withNumberOfRegisters @IntLength @Auto @(BaseField c) $ withIsoFieldElemUInt @IntLength @c $
+  withGetRegisterSize @IntLength @Auto @(BaseField c) $ withCeilRegSize @(GetRegisterSize (BaseField c) IntLength Auto) @OrdWord $
+    fromConstant (\(VarByteString l b) i@(Int t) ->
+      let fr = bool (error "index is not in interval") (from t) (isNotNegative i && t Symbolic.< fromConstant (value @BSLength))
```
TurtlePU, line 509: This is going to throw an error instantly when evaluated into circuit

TurtlePU: Instead of this, you have to return a `Symbolic.nothing` in case when an index is out of bounds

symbolic-base/src/ZkFold/Symbolic/Data/VarByteString.hs
```diff
@@ -361,3 +364,9 @@ shiftWordsR (Words regs) p2
             s <- newAssigned $ \p ->  p h + scale ((2 :: Natural) ^ (regSize -! remShift)) (p carry)
             pure (s : acc, l)
 
+dropZeros :: forall n m c . (Symbolic c, KnownNat n, n <= m, KnownNat (m - n)) => VarByteString m c -> VarByteString n c
+dropZeros VarByteString{..} = ifThenElse (bsLength < feN) bsNMoreLen bsNLessLen
```
TurtlePU, line 368: Sorry, I don't understand what this function does, can you provide a comment?

hovanja2011: If the composition of two strings is shorter than the maximum length of the type, then the zero bits on the left are removed (since the strings are filled on the right, see the 'append' function for VarByteString). 
If the composition of two strings is longer than the size of our type, then the first n bits are kept, and the rest are cut off.

TurtlePU: Thanks for the clarification!
(I meant adding it as a comment to the function, but this works too, thank you)

TurtlePU (approved):

PR #511, opened 2025-03-06
Use set inclusion constraints in bitwise operations

hovanja2011:


symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
```diff
@@ -143,9 +143,13 @@ acSizeN = length . acSystem
 acSizeM :: ArithmeticCircuit a p i o -> Natural
 acSizeM = length . acWitness
 
--- | Calculates the number of range lookups in the system.
-acSizeR :: ArithmeticCircuit a p i o -> Natural
-acSizeR = sum . map length . M.elems . acLookup
+-- | Calculates the number of all lookups in the system.
```
TurtlePU, line 146: I think we need to concretize the documentation here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit.hs
```diff
@@ -143,9 +143,13 @@ acSizeN = length . acSystem
...
+acSizeL :: ArithmeticCircuit a p i o -> Natural
+acSizeL = sum . map length . M.elems . acLookup
+
+-- | Calculates the number of all lookups in the system.
```
TurtlePU, line 150: I think we need to concretize the documentation here

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Var.hs
```diff
@@ -53,6 +55,10 @@ data Var a i
 toVar :: Semiring a => SysVar i -> Var a i
 toVar x = LinVar one x zero
 
+fromVar :: Arithmetic a => Var a i -> SysVar i
```
TurtlePU, line 58: I think we should either return a `Maybe (SysVar i)` here or remove this function altogether because most probably, a user _will_ encounter variables which are not SysVar-s and they will have to handle it somehow

TurtlePU: All in all, I guess, we need to store `SysVar`-s in the lookup field of circuit in the end

symbolic-base/src/ZkFold/Symbolic/Data/Bool.hs
TurtlePU, line 1: `Bool` doesn't get any better optimized from lookup constraints, we do not have to change it

symbolic-base/src/ZkFold/Symbolic/Data/ByteString.hs
TurtlePU, line 1: Once again, we wouldn't win anything from doing bitwise operations bit-by-bit through lookups

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Var.hs
vlasin: I guess to handle it properly, we should add a new `SysVar` whenever we add a lookup constraint on a `Var` unless it is already `SysVar`. Is that what is going on?

symbolic-base/src/ZkFold/Symbolic/Data/ByteString.hs
TurtlePU: What we need to do here actually is to store data in a `ByteString` in chunks of some size (for now we can set it to be 16 and later generalize to how it's done in `UInt`), where each chunk is stored in a single circuit variable. Then you need to define an operation which works like this:
```
f(b0 + 2*b1 + ... + 2^15*b15) = b0 + 4*b1 + ... + 4^15*b15
```
After doing so, note that
```
f(x) + f(y) = x xor y + 2 * x and y
```
Where `xor` is a bitwise xor of chunks and `and` is a bitwise and of chunks. From there, the rest of boolean operations easily follow

symbolic-base/src/ZkFold/Symbolic/Data/Lookup.hs
```diff
@@ -0,0 +1,49 @@
...
+import           ZkFold.Symbolic.MonadCircuit
+
+
+binLookup :: Arithmetic a => LookupTable a Par1
+binLookup = Ranges $ S.singleton (zero, one)
+
+bin2Lookup :: Arithmetic a => LookupTable a (Par1 :*: Par1)
+bin2Lookup = Product binLookup binLookup
+
+newBinLookup :: forall f var a w m.
```
TurtlePU, line 27: Nice!

symbolic-examples/stats/ByteString.And.32
```diff
@@ -1,4 +1,5 @@
 ByteString.And.32
-Number of constraints: 287
+Number of constraints: 254
 Number of variables: 286
-Number of range lookups: 0
\ No newline at end of file
+Number of all lookups: 0
```
TurtlePU, line 4: How is it possible that `and` as defined above doesn't introduce any range lookups?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Var.hs
TurtlePU: It should be like this, but it's not what is happening right now here

vlasin:
@TurtlePU, could you please fix the remaining issues, if any?

symbolic-base/src/ZkFold/Symbolic/Data/ByteString.hs
TurtlePU: Will be done in the next PR

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Internal.hs
```diff
@@ -84,6 +86,9 @@ import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.Witness
 import           ZkFold.Symbolic.Compiler.ArithmeticCircuit.WitnessEstimation
 import           ZkFold.Symbolic.Fold
 import           ZkFold.Symbolic.MonadCircuit
+import Data.Typeable (Typeable)
```
vlasin, line 89: No formatting

vlasin (dismissed):

PR #515, opened 2025-03-07
feat(#509): refactor ledger value type to map from currency symbol to…

### LINKED ISSUE #509: Refactor `Value` type to represent map from currency symbol to pairs of token name and amount

sourabhxyz:
… another map from token name to amount

Closes #509.


symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
```diff
@@ -53,19 +62,31 @@ addValue ::
...
+       (UnsafeMultiAssetValue (singleton (mintingPolicy, singleton (tokenInstance, tokenQuantity))))
+       multiVal
+  where
+    addTokenAmount tokenAmountList =
+      let (tokenHead@(tokenHeadToken, tokenHeadAmount), tokenTail) = uncons tokenAmountList
+          tokenAmountAdded =
+            ifThenElse (tokenHeadToken == tokenInstance)
+              ((tokenHeadToken, tokenHeadAmount + tokenQuantity) .: tokenTail)
+              (tokenHead .: addTokenAmount tokenTail)
+      in ifThenElse (null tokenAmountList)
```
sourabhxyz, line 87: @vlasin If `null tokenAmountList` is true, would `tokenAmountAdded` be exercised?

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
```diff
@@ -35,12 +37,19 @@ data Value context = Value
   , tokenQuantity :: UInt 64 Auto context
   }
 
-newtype MultiAssetValue context = UnsafeMultiAssetValue (List context (Value context))
+-- | Denotes multiple values.
+newtype MultiAssetValue context = UnsafeMultiAssetValue (List context (CurrencySymbol context, List context (Token context, UInt 64 Auto context)))
 
+-- | Constraint for a tuple to represent a symbolic data.
+type TupleSymbolicData c x y = (SymbolicData (x c), SymbolicData (y c), HApplicative (Context (x c)), Context (x c) ~ Context (y c), Support (x c) ~ Support (y c))
```
sourabhxyz, line 44: I should likely give this a better name and move it in more suitable module.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
vlasin: Yes

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
vlasin: We have an instance of `SymbolicData` for a pair. This should not be necessary.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
```diff
@@ -26,62 +51,106 @@ type MintingContract tx w context = Contract tx (Token context) w context
...
-  => Conditional (Bool context) (MultiAssetValue context)
-  => BooleanOf (Token context) ~ Bool context
-  => Eq (Token context)
-  => SymbolicEq (CurrencySymbol context)
-  => Context (CurrencySymbol context) ~ context
+addValue Value {..} (UnsafeMultiAssetValue valList) =
+  let (policyExisted, _, _, _, r) =
+        Symbolic.List.foldr (Morph \((yp :: CurrencySymbol s, yas :: List s ((Token s, Amount s))), (found :: Bool s, mintingPolicy' :: CurrencySymbol s, tokenInstance' :: Token s, tokenQuantity' :: Amount s, ys)) ->
+          let isSame :: Bool s = mintingPolicy' == yp
+              tokenAmountAdded = addTokenAmount tokenInstance' tokenQuantity' yas
```
sourabhxyz, line 132: @vlasin @TurtlePU Currently this computation occurs for all currency symbol entries in list, is there a way to only have it computed in case `isSame` is true?

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
sourabhxyz: Earlier types `Token context` etc. were not having required instances and that's why I needed this constraint. But now that I have provided those instances, this is no longer needed.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
vlasin: Well, in general, all symbolic computations are "strict", but this fact does not rule out the possibility of some optimizations.

The reason why there is no easy way to avoid doing all the extra computations (on each iteration of `foldr`) is because `isSame` is a "variable" when we compile the circuit (i.e., executing this code with `ArithmeticCircuit` context). It does not have a particular "value" until we plug concrete numbers into the already-built circuit.

vlasin (approved):

PR #517, opened 2025-03-09
Poly hierarchy

diS3e:


symbolic-base/src/ZkFold/Base/Protocol/IVC/Internal.hs
```diff
@@ -86,6 +86,7 @@ type IVCAssumptions ctx0 ctx1 algo d k a i p c f =
     , RecursiveFunctionAssumptions algo d a i c (FieldElement ctx0) ctx0
     , ctx1 ~ ArithmeticCircuit a (RecursiveI i :*: RecursiveP d k i p c) U1
     , RecursiveFunctionAssumptions algo d a i c (FieldElement ctx1) ctx1
+    , Eq (ArithmeticCircuit a (RecursiveI i :*: RecursiveP d 1 i p c) U1 Par1)
```
TurtlePU, line 89: Why is this constraint needed?

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -67,28 +44,88 @@ infixl 6 .+, +.
 
 -------------------------------- Arbitrary degree polynomials --------------------------------
 
--- TODO (Issue #17): hide constructor
-newtype Poly c = P (V.Vector c)
-    deriving (Eq, Show, Functor, Generic, NFData)
+class
+    ( Ring c
+    , AdditiveGroup (poly c)
+    ) => UnivariateRingPolynomial poly c where
```
TurtlePU, line 50: As it does not link different instantiations of `poly` together (`poly c` and `poly d` for different `c` and `d` in a single function signature), I think it's better to set `poly` to be a `Type` rather than `Type -> Type`, like this:
```haskell
-- | A "main" type should go last to allow deriving.
-- A functional dependency wouldn't hurt.
class (Ring c, AdditiveGroup poly) => UnivariateRingPolynomial c poly | poly -> c where
  constant :: c -> poly -- actually can be just a `FromConstant`
  fromPoly :: poly -> V.Vector c
  ...
```

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -67,28 +44,88 @@ infixl 6 .+, +.
...
 
-constant :: c -> Poly c
-constant = P . V.singleton
+    fromPoly :: poly c -> V.Vector c
 
--- | A polynomial of form cx^d
---
-monomial :: Ring c => Natural -> c -> Poly c
-monomial d c = P $ V.fromList (replicate d zero P.<> [c])
+    lt :: poly c -> c
```
TurtlePU, line 56: A doc comment wouldn't hurt here

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -67,28 +44,88 @@ infixl 6 .+, +.
...
+    deg :: poly c -> Integer
 
-instance {-# OVERLAPPING #-} FromConstant (Poly c) (Poly c)
+    toPoly :: V.Vector c -> poly c
+
+    monomial :: Natural -> c -> poly c
+
+    evalPoly :: poly c -> c -> c
+
+    scaleP :: c -> Natural -> poly c -> poly c
```
TurtlePU, line 67: A comment wouldn't hurt here

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -251,57 +288,143 @@ instance (Field c, Eq c) => MultiplicativeMonoid (Poly c) where
...
-fromPolyVec (PV cs) = cs
+    -- -- p(x) = a0
+    polyVecConstant :: c -> pv c size
 
-rewrapPolyVec :: (V.Vector c -> V.Vector c) -> PolyVec c size -> PolyVec c size
-rewrapPolyVec f (PV x) = PV (f x)
+    -- p(x) = a1 * x + a0
+    polyVecLinear :: c -> c -> pv c size
+
+    -- p(x) = a2 * x^2 + a1 * x + a0
```
TurtlePU, line 316: In which order are `a2, a1, a0` supplied to the function? Is it like `@polyVecQuadratic a2 a1 a0 = a2 * x^2 + a1 * x + a0@` or the other way around?

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Univariate.hs
```diff
@@ -251,57 +288,143 @@ instance (Field c, Eq c) => MultiplicativeMonoid (Poly c) where
...
+newtype PolyVec c (size :: Natural) = PV (V.Vector c)
+    deriving (Eq, Show, Generic, NFData)
 
-scaleP :: Ring c => c -> Natural -> Poly c -> Poly c
-scaleP a n (P cs) = P $ V.replicate (fromIntegral n) zero V.++ fmap (a *) cs
+class
+    ( Ring c
+    , KnownNat size
+    , AdditiveGroup (pv c size)
+    ) => UnivariateRingPolyVec pv c size where
```
TurtlePU, line 300: Once again &mdash; what is the benefit of making `pv` of kind `Type -> Natural -> Type` rather than just `Type`?

diS3e: We need at least a `Natural -> Type` kind because of `polyVecInLagrangeBasis`

TurtlePU: I see
Then I'd consider dropping the `size` parameter from typeclass declaration:
```haskell
class (Ring c, forall size. KnownNat size => AdditiveGroup (pv size))
    => UnivariateRingPolyVec c pv where
  toPolyVec :: KnownNat size => V.Vector c -> pv size
  poly2vec :: (UnivariateRingPolynomial c poly, KnownNat size) => poly -> pv size
  ...

class (Field c, UnivariateRingPolyVec c pv, forall size. KnownNat size => Ring (pv size))
    => UnivariateFieldPolyVec pv c where
  (.*.) :: pv size -> pv size -> pv size
  polyVecInLagrangeBasis :: (KnownNat n, KnownNat size) => c -> pv n -> pv size
  ...
```

vlasin:
:fire:

symbolic-base/src/ZkFold/Base/Algebra/Basic/Class.hs
```diff
@@ -93,6 +93,9 @@ class Scale b a where
     default scale :: (FromConstant b a, MultiplicativeSemigroup a) => b -> a -> a
     scale = (*) . fromConstant
 
+class MultiScale p s g where
```
TurtlePU, line 96: I think a fundep `| p s -> g` can be added here to improve type inference a bit

TurtlePU: (Also maybe I'd rename it to something like "bilinear" as this is not really a scaling operation as it changes the type of the result)

symbolic-base/src/ZkFold/Base/Algebra/Basic/Field.hs
```diff
@@ -226,48 +226,51 @@ instance Field f => AdditiveGroup (Ext2 f e) where
...
     rootOfUnity n = (\r -> Ext2 r zero) <$> rootOfUnity n
 
-instance (FromConstant f f', Field f') => FromConstant f (Ext2 f' e) where
-    fromConstant e = Ext2 (fromConstant e) zero
+instance {-# OVERLAPPING #-} (Field f, IrreduciblePoly poly f e) => FromConstant Natural (Ext2 f e) where
+    fromConstant = flip scale one
 
-instance {-# OVERLAPPING #-} (Field f, Eq f, IrreduciblePoly f e) => FromConstant (Poly f) (Ext2 f e) where
-    fromConstant p = case fromPoly . snd $ qr p (irreduciblePoly @f @e) of
+instance {-# OVERLAPPING #-} (Field f, IrreduciblePoly poly f e) => FromConstant Integer (Ext2 f e) where
```
TurtlePU, line 256: I think the `FromConstant Natural` and `FromConstant Integer` instances can be removed since we have a `FromConstant c` instance which covers its cases

symbolic-base/src/ZkFold/Base/Algebra/Basic/Field.hs
```diff
@@ -319,23 +322,26 @@ instance (Field f, Eq f, IrreduciblePoly f e) => Field (Ext3 f e) where
 
     rootOfUnity n = (\r -> Ext3 r zero zero) <$> rootOfUnity n
 
-instance (FromConstant f f', Field f') => FromConstant f (Ext3 f' ip) where
-    fromConstant e = Ext3 (fromConstant e) zero zero
+instance {-# OVERLAPPING #-} (Field f, IrreduciblePoly poly f e) => FromConstant Natural (Ext3 f e) where
+    fromConstant = flip scale one
+
+instance {-# OVERLAPPING #-} (Field f, IrreduciblePoly poly f e) => FromConstant Integer (Ext3 f e) where
```
TurtlePU, line 328: Same here -- the `FromConstant c` instance covers this functionality

symbolic-base/src/ZkFold/Base/Algebra/Basic/Class.hs
vlasin: Hmm, I am thinking how about adding a class constraint (`Scale`) and a default implementation, instead?

diS3e: > I think a fundep `| p s -> g` can be added here to improve type inference a bit

Maybe we can specify the dependence of the result on the argument like this?
```haskell
class Container p where
    type Content p :: Type

class 
    ( Container s
    , Container p
    ) => MultiScale p s where

    mscale :: p -> s -> Content p
    default mscale :: (p ~ Content p, Scale s p) => p -> s -> Content p
    mscale = Haskell.flip scale

```

TurtlePU: > Maybe we can specify the dependence of the result on the argument like this?

I'd advise against doing so, because there is no inherent mathematical meaning behind introducing a typeclass other than constraining the class so that it is closer to what we have in Plonk. But, as for me, I cannot guess in advance -- maybe, we will need a similar abstraction somewhere else, where these `Container` constraints wouldn't make sense?

TurtlePU: Also, if we reformulate Plonk in the same way how it's done with IVC (using records of functions instead of class instences), we needn't introduce a MultiScale class at all -- we can just pass a function responsible for the bilinear operation as an argument of this record constructor

TurtlePU (approved):

vlasin (approved):

PR #518, opened 2025-03-11
Vks4git/wallet m4

vks4git:


symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
TurtlePU, line 1: Looks domain-specific to me, maybe it would be better if it belonged in `symbolic-cardano`?
(Or one could say that this whole JWT stuff deserves to be placed in its own little library package)

symbolic-base/src/ZkFold/Symbolic/Data/JWT/RS256.hs
```diff
@@ -0,0 +1,84 @@
...
+deriving instance
+    ( P.Show (PublicKey 2048 ctx)
+    , P.Show (VarByteString 320 ctx)
+    ) => P.Show (Certificate ctx)
+deriving instance
+    ( NFData (PublicKey 2048 ctx)
+    , NFData (VarByteString 320 ctx)
+    ) => NFData (Certificate ctx)
+instance
+    ( SymbolicData (PublicKey 2048 ctx)
```
TurtlePU, line 41: Why is this constraint needed?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/RS256.hs
```diff
@@ -0,0 +1,84 @@
...
+    , Symbolic ctx
+    ) => SymbolicData (Certificate ctx)
+instance
+    ( SymbolicInput (PublicKey 2048 ctx)
```
TurtlePU, line 45: Why is this constraint needed?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/RS256.hs
```diff
@@ -0,0 +1,84 @@
...
+deriving instance
+    ( P.Show (PrivateKey 2048 ctx)
+    , P.Show (VarByteString 320 ctx)
+    ) => P.Show (SigningKey ctx)
+deriving instance
+    ( NFData (PrivateKey 2048 ctx)
+    , NFData (VarByteString 320 ctx)
+    ) => NFData (SigningKey ctx)
+instance
+    ( SymbolicData (PrivateKey 2048 ctx)
```
TurtlePU, line 72: Why is this constraint needed?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/RS256.hs
```diff
@@ -0,0 +1,84 @@
...
+    , Symbolic ctx
+    ) => SymbolicData (SigningKey ctx)
+instance
+    ( SymbolicInput (PrivateKey 2048 ctx)
```
TurtlePU, line 76: Why is this constraint needed?

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
```diff
@@ -5,123 +5,73 @@
...
--- | RSA Private key with Key ID
+-- | Signing algorithm for JWT (such as RS256)
 --
-data SigningKey ctx
-    = SigningKey
-        { prvKid :: VarByteString 320 ctx
-        , prvKey :: PrivateKey 2048 ctx
-        }
-    deriving Generic
+class SigningAlgorithm (alg :: Symbol) where
```
TurtlePU, line 55: So in essence, a `SigningAlgorithm` is a dict of types. With `DataKinds` and `TypeFamilies`, you can declare it as such:
```haskell
data SigningAlgorithm = MkSigningAlgorithm
  { saSKey :: Type
  , saVKey :: Type
  , saSignature :: Type
  , saHash :: Type
  }

type family SKey (_ :: SigningAlgorithm) :: Type where SKey (MkSigningAlgorithm skey _ _ _) = skey
...

type RS256 ctx = MkSigningAlgorithm (SigningKey ctx) (Certificate ctx) (ByteString 2048 ctx) (ByteString 256 ctx)
```

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
```diff
@@ -0,0 +1,138 @@
...
+    ) => P.Show (TwitchPayload ctx)
+deriving instance Symbolic ctx => SymbolicData (TwitchPayload ctx)
+deriving instance Symbolic ctx => SymbolicInput (TwitchPayload ctx)
+instance Symbolic ctx => Arbitrary (TwitchPayload ctx) where
+    arbitrary = genericArbitrary uniform
+
+instance Symbolic ctx => FromJSON (TwitchPayload ctx) where
+    parseJSON = genericParseJSON (aesonPrefix snakeCase) . stringify
+
+instance (Symbolic ctx, Context (TwitchPayload ctx) ~ ctx) => IsSymbolicJSON (TwitchPayload ctx) where
```
TurtlePU, line 109: Why is the `Context (TwitchPayload ctx) ~ ctx` constraint needed here?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
```diff
@@ -0,0 +1,138 @@
...
+
+instance
+  ( Symbolic ctx
+  , NFData (ctx (V.Vector 8))
+  , NFData (ctx (V.Vector 9056))
+  ) => IsBits (TwitchPayload ctx) where
+    type BitCount (TwitchPayload ctx) = 9056
+    toBits = toAsciiBits
+
+instance (Symbolic ctx, TokenBits (TwitchPayload ctx), RSA.RSA 2048 10328 ctx) => IsTokenPayload "RS256" (TwitchPayload ctx) where
```
TurtlePU, line 129: Why is the `TokenBits (TwitchPayload ctx)` constraint needed here?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Utils.hs
```diff
@@ -0,0 +1,245 @@
...
+
+mulMod :: forall n . Dict (Mod (Next6 n) 6 ~ 0)
+mulMod = unsafeAxiom
+
+withDivMul :: forall a b {r}. (Mod a b ~ 0) => ((Div a b) * b ~ a => r) -> r
+withDivMul = withDict (divMul @a @b)
+
+---------------------------------------------------------------------------------------------------
+
+feToUInt :: forall n ctx . Symbolic ctx => FieldElement ctx -> UInt (BufLen n) ('Fixed (BufLen n)) ctx
```
TurtlePU, line 114: Maybe we could move this function to `UInt` module

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Utils.hs
```diff
@@ -0,0 +1,245 @@
...
+feToUInt (FieldElement c) = UInt $ withDict (oneReg @n @ctx) $ hmap (V.singleton . unPar1) c
+
+uintToFe :: forall n ctx . Symbolic ctx => UInt (BufLen n) ('Fixed (BufLen n)) ctx -> FieldElement ctx
```
TurtlePU, line 117: Maybe we could move this function to `UInt` module

symbolic-base/src/ZkFold/Symbolic/Data/JWT/RS256.hs
vks4git: To replace two constraints of the type `KnownNat (NumberOfRegisters ...)`. In every other instance where SymbolicData is a constraint the reasoning is the same

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
vks4git: Because `IsSymbolicJSON a` has a function from `a` to `VarByteString n (Context a)`. It does not know that `ctx` is, in fact, `Context (TwitchPayload ctx)`

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
vks4git: What advantage will it provide? From the user's perspective, the only thing which will change will be calling `signPayload @(RS256 ctx)` instead of `signPayload @"RS256"`, right?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
vks4git: because both `signPayload` and `verifyJWT` use `tokenBits` from this class

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
vks4git: But it's not related to Cardano, it's just an interface for verifying JWTs. The contract which uses these tokens is indeed [in symbolic-cardano](https://github.com/zkFold/symbolic/blob/main/symbolic-cardano/src/ZkFold/Symbolic/Cardano/Contracts/ZkLogin.hs), though

symbolic-base/src/ZkFold/Symbolic/Data/JWT.hs
TurtlePU: Yes, that's right
Also this would allow to create new signing algorithms inplace (though I don't know if it's useful)
It's just that I'm not in favor of introducing typeclasses where some other mechanism would suffice, but you're right that here the difference is really minor

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
TurtlePU: That's really strange that GHC cannot compute `Context (TwitchPayload ctx)` and unify it with `ctx`, could you please provide the error it reports here for me so I could take a look?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
TurtlePU: Oh, ok, I mean shouldn't this constraint be inferrable by itself? Or is it that the needed instance is currently missing?

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
vks4git: Oh wow, GHC actually did infer the type equality! But I saw this error somewhere, I swear... Thank you!

symbolic-base/src/ZkFold/Symbolic/Data/JWT/Twitch.hs
vks4git: Oh, no, TokenBits also includes several NFData and KnownNat constraints. If you remove it, GHC complains that
```src/ZkFold/Symbolic/Data/JWT/Twitch.hs:137:56: error: [GHC-39999]
    • Could not deduce ‘NFData (ctx (V.Vector 8))’
        arising from a use of ‘tokenBits’
```

TurtlePU: Oh, ok, understood!

TurtlePU (approved):

vlasin (approved):

PR #520, opened 2025-03-12
UInt OrdWord removal

### LINKED ISSUE #486: Simplify Haskell constraints for `instance Ord (UInt n r c)`

echatav:
Resolves #486


symbolic-base/src/ZkFold/Symbolic/Data/UInt.hs
```diff
@@ -447,14 +443,14 @@ instance ( Symbolic c, KnownNat n, KnownRegisterSize r
     x <  y = y > x
 
     (UInt u1) >= (UInt u2) =
-        let w1 = asWords @OrdWord @regSize u1
-            w2 = asWords @OrdWord @regSize u2
-         in bitwiseGE @OrdWord w1 w2
+        let w1 = asWords @regSize u1
+            w2 = asWords @regSize u2
+         in bitwiseGE @regSize w1 w2
```
vks4git, line 448: Have you also changed the implementation of `bitwiseGE`? Invoking it with word size as big as the whole register will generate impossible range constraints.

PR #528, opened 2025-03-20
Build symbolic with both GHC 9.6 and GHC 9.12

vks4git:


vks4git:
Converted to draft since stylish-haskell removed the changes I introduced...

symbolic-base/src/ZkFold/Base/Algebra/Polynomials/Multivariate/Polynomial.hs
```diff
@@ -8,10 +8,7 @@ import           Control.DeepSeq                                       (NFData)
 import           Data.Aeson                                            (FromJSON, ToJSON)
 import           Data.Bifunctor                                        (Bifunctor (..))
 import           Data.Functor                                          ((<&>))
-import           Data.List                                             (intercalate)
-#if __GLASGOW_HASKELL__ < 912
-import           Data.List                                             (foldl')
-#endif
+import           Data.List                                             (foldl', intercalate)
```
TurtlePU, line 11: I think that, actually, what stylish-haskell did here is ok because for both GHC 9.6 and 9.12, foldl' can be found here

vks4git: But then GHC 9.12 will complain about redundant imports, and we're building with -Werror

TurtlePU: Oof, ok
Technically, we could rewrite it to not depend on the Prelude (or, at least, list the dependencies explicitly)

TurtlePU:
I guess we can close this since #548 subsumes it?

PR #533, opened 2025-03-24
feat(#532): `Conditional`, `Eq` instance for `Hash` type

### LINKED ISSUE #532: [symbolic-base] `Conditional`, `Eq` instance for `Hash` type

sourabhxyz:
Closes #532.

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/Hash.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/Hash.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module ZkFold.Symbolic.Data.Hash where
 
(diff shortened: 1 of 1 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/Hash.hs
```diff
@@ -38,6 +40,8 @@ data Hash h a = Hash
 
 instance (SymbolicOutput h, SymbolicOutput a) => SymbolicData (Hash h a)
 instance (SymbolicInput h, SymbolicInput a) => SymbolicInput (Hash h a)
+instance (c ~ (Context h), Conditional (Bool c) h, Symbolic c, SymbolicData a) => Conditional (Bool c) (Hash h a)
```
TurtlePU, line 43: The `Symbolic c` constraint can be inferred from `c ~ Context h`, so it can be removed

sourabhxyz: @TurtlePU unfortunately, I get following error without it,
```
Could not deduce ‘Symbolic c’
    arising from a use of ‘ZkFold.Symbolic.Data.Conditional.$dmbool’
  from the context: (c ~ Context h, Conditional (Bool c) h,
                     SymbolicData a)
```

symbolic-base/src/ZkFold/Symbolic/Data/Hash.hs
```diff
@@ -38,6 +40,8 @@ data Hash h a = Hash
...
+instance (c ~ (Context h), Symbolic c, SymbolicData a, BooleanOf h ~ Bool c, Eq h) => Eq (Hash h a)
```
TurtlePU, line 44: The `Symbolic c` constraint can be inferred from `c ~ Context h`, so it can be removed
Also `c` can be substituted into `Bool c`, removing its mention completely:
```haskell
(SymbolicData a, Eq h, BooleanOf h ~ Bool (Context h))
```

symbolic-base/src/ZkFold/Symbolic/Data/Hash.hs
TurtlePU: Oh, sorry, thought first that there's also a `SymbolicData h` constraint
Since there's nothing like it, looks like they constraints are ok

PR #546, opened 2025-04-02
feat: avoid `Foldable` constraint on symbolic list type

sourabhxyz:


symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types.hs
```diff
@@ -5,27 +5,16 @@ module ZkFold.Symbolic.Ledger.Types (
     module ZkFold.Symbolic.Ledger.Types.Transaction,
     module ZkFold.Symbolic.Ledger.Types.Root,
     module ZkFold.Symbolic.Ledger.Types.Datum,
+    module ZkFold.Symbolic.Ledger.Types.Value,
```
sourabhxyz, line 8: Missed to re-export it previously.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Value.hs
```diff
@@ -115,9 +115,8 @@ addAssetValue givenAssetVal (UnsafeAssetValues assetValList) =
 -- | Safe constructor for 'AssetValues'.
 assetValuesFromList ::
      SymbolicFold context
-  => Foldable (List context)
   => KnownRegistersAssetQuantity context
   => List context (AssetValue context)
   -> AssetValues context
-assetValuesFromList = foldr addAssetValue emptyAssetValues
+assetValuesFromList = Symbolic.List.foldr (Morph \(x, acc) -> addAssetValue x acc) emptyAssetValues
```
sourabhxyz, line 121: Using `foldr` instead of `foldl` because I think it would be better if generated list is in sync with the order of values given, though not strictly required.

vlasin (approved):

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types.hs
```diff
@@ -48,45 +37,9 @@ import           ZkFold.Symbolic.Ledger.Types.Value
...
-
-    , SymbolicOutput (Transaction context)
-    , Context (Transaction context) ~ context
-
-    , SymbolicOutput (Output context)
-    , Context (Output context) ~ context
+    ( KnownRegistersAssetQuantity context
+    , KnownRegistersOutputIndex context
+    , KnownRegisters context 11 Auto
+    , Symbolic context
```
TurtlePU, line 43: btw `SymbolicFold` implies `Symbolic`, so this constraint is also not needed

sourabhxyz: Thanks! Would omit it in https://github.com/zkFold/symbolic/pull/547.

PR #560, opened 2025-04-14
feat(#537): tx validation

sourabhxyz:


symbolic-ledger/src/ZkFold/Symbolic/Ledger/Validation/Transaction/Core.hs
```diff
@@ -1,75 +1,117 @@
...
+import           ZkFold.Symbolic.Data.List        (List)
+import           ZkFold.Symbolic.Data.Maybe       (Maybe)
 import           ZkFold.Symbolic.Data.Morph
 import           ZkFold.Symbolic.Ledger.Types
 
+{- | Common witness for 'Input's validation, i.e., to verify that input belongs to valid UTxO set.
+
+__Note__: Having it as a type synonym helps significantly with compilation times.
+-}
+type UtxoWitness context =
```
sourabhxyz, line 28: @vlasin @TurtlePU Apparently `UtxoWitness` is causing compilation time blowup. I have made simplification in where I require single `UtxoWitness` for entire `TransactionBatch` (instead of requiring witness for each input inside each `Transaction` inside each `TransactionBatchData` inside each `TransactionBatch`) thus this `UtxoWitness` isn't further nested in a `List`.
But even in it's current format & usage, it takes significant time to compile (around 10 mins on my machine to solely compile validation related modules of `symbolic-ledger`).

TurtlePU: I guess this might happen because `List` usage generates a lot of constraints which end up being chained.
Also, why is `ImpredicativeTypes` enabled? Is it needed anywhere here? This might lead to the slowdown you experience as well

TurtlePU: About chained constraints -- I think removing the `context` param from `List` declaration might help

TurtlePU: Same question about `UndecidableInstances` -- is it really required here?

TurtlePU: Oh, I see! Then there's two ways to remove Impredicative Types (either one will do):
* add corresponding instances, as you suggested
* simplify Symbolic so that context switching is not needed

Both are right things to do and both will be done, but I guess the first solution is simpler to implement right now

sourabhxyz:
Superseded by https://github.com/zkFold/symbolic/pull/696.

PR #564, opened 2025-04-16
Test refactor

vlasin:


symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+  => x c
+
+specSymbolicFunction0 :: forall a x o .
+  ( Show a
+  , Eq (x (Interpreter a))
+  , Show (x (Interpreter a))
+  , MatchingSymbolicInput x o (Interpreter a) (ArithmeticCircuit a U1)
+  ) => String -> SymbolicFunction0 x -> Spec
+specSymbolicFunction0 desc v = describe desc $ do
+  it "evaluates correctly" $ evalCircuit0 @a (compileCircuit0 v) === v
```
TurtlePU, line 173: This property is always satisfied because each of the contexts evaluates the same witnesses in the same way thanks to the way `Symbolic` interface is defined

symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+
+{-
+  For all symbolic types we need to do the following:
+  1. Check that `fromConstant` and `toConstant` are inverse functions.
+  2. For all related functions, check that the circuit evaluation is equivalent to the function evaluation and
+     that the circuit's constraints are satisfied by the respective witness.
+  3. Run all type-specific checks in the interpreter context.
+-}
+
+specConstantRoundtrip :: forall a x .
```
TurtlePU, line 42: Since `x` is always used here with `Interpreter a`, the type can be generalized a bit:
```haskell
specConstantRoundTrip :: forall x .
    ( Arbitrary x, Eq x, Eq (Const x), Show x, Show (Const x)
    , FromConstant (Const x) x, ToConstant x
    ) => String -> String -> Gen (Const x) -> Spec
```

symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+  , Context (y c') ~ c'
+  , Layout (y c) ~ Layout (y c')
+  , Layout (y c) ~ o
+  , Payload (y c) ~ Payload (y c')
+  , Payload (y c) ~ U1
+  , BaseField (Context (y c)) ~ BaseField (Context (y c'))
+  , Binary (BaseField (Context (y c)))
+  )
+
+evalCircuit0 :: forall a x o . (Arithmetic a, Binary a, Representable o
```
TurtlePU, line 84: Once again, `x (Interpreter a)` can be generalized away here

symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+  , Context (x (Interpreter a)) ~ Interpreter a
+  , Layout (x (Interpreter a)) ~ ix
+  , SymbolicInput (y (Interpreter a))
+  , Context (y (Interpreter a)) ~ Interpreter a
+  , Layout (y (Interpreter a)) ~ iy
+  , SymbolicOutput (z (Interpreter a))
+  , Context (z (Interpreter a)) ~ Interpreter a
+  , Layout (z (Interpreter a)) ~ o
+  , Payload (z (Interpreter a)) ~ U1
+  , i ~ ix :*: iy
```
TurtlePU, line 116: `i` can be inlined into the argument type

symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+  , ToConstant (x (Interpreter a))
+  ) => String -> String -> Gen (Const (x (Interpreter a))) -> Spec
+specConstantRoundtrip symType hType gen = do
+  it (symType ++ " embeds " ++ hType) $
+    \(x :: x (Interpreter a)) -> fromConstant (toConstant x :: Const (x (Interpreter a))) === x
+  it (hType ++ " embeds " ++ symType) $ do
+    (x :: Const (x (Interpreter a))) <- gen
+    return $ toConstant (fromConstant x :: x (Interpreter a)) === x
+
+type MatchingSymbolicInput x i c c' =
```
TurtlePU, line 58: What about this?
```
type Match constr a i l x x' =
    ( constr x, constr x', Context x ~ Interpreter a, Context x' ~ ArithmeticCircuit a i
    , Layout x ~ Layout x', Layout x ~ l, Payload x ~ Payload x', Payload x ~ U1
    )
type Matching constr a i l x = Match constr a i l (x (Interpreter a)) (x (ArithmeticCircuit a i))
-- type MatchingSymbolicInput a i l x = Matching SymbolicInput a i l x
-- type MatchingSymbolicOutput a i l x = Matching SymbolicOutput a i l x
```

symbolic-base/test/Tests/Symbolic/Data/Common.hs
```diff
@@ -0,0 +1,234 @@
...
+  , Payload (x (ArithmeticCircuit a i)) ~ U1
+  , SymbolicInput (y (ArithmeticCircuit a i))
+  , Context (y (ArithmeticCircuit a i)) ~ ArithmeticCircuit a i
+  , Layout (y (ArithmeticCircuit a i)) ~ iy
+  , Payload (y (ArithmeticCircuit a i)) ~ U1
+  , SymbolicOutput (z (ArithmeticCircuit a i))
+  , Context (z (ArithmeticCircuit a i)) ~ ArithmeticCircuit a i
+  , Layout (z (ArithmeticCircuit a i)) ~ o
+  , i ~ ix :*: iy
+  ) => (forall c . Symbolic c => x c -> y c -> z c) -> ArithmeticCircuit a i o
```
TurtlePU, line 159: If we specialize `c` with `ArithmeticCircuit a i` here, we can 1) get rid of explicit specialization (`func @(...)`) next 2) generalize `z (ArithmeticCircuit a i)` and the like away

symbolic-base/test/Tests/Symbolic/Data/Common.hs
TurtlePU: It then suffices to define
```haskell
specConstantRoundtrip = specConstantRoundtrip'
```

TurtlePU (approved):

PR #565, opened 2025-04-17
CI update

vlasin:


vlasin:
I had to comment out one benchmark that was not passing. The `bench-compiler-performance` fails because `productMod` and `expMod` for `UInt` exit with a "division by zero" error.

@vks4git

TurtlePU (approved):

PR #570, opened 2025-04-21
PlonkUp update: lookups

vlasin:
Adding support for 3-column lookups.


symbolic-base/src/ZkFold/Prelude.hs
```diff
@@ -97,9 +103,8 @@ assert statement obj x = if statement then x else error $ show obj
 chooseNatural :: (Natural, Natural) -> Gen Natural
 chooseNatural (lo, hi) = integerToNatural <$> chooseInteger (fromIntegral lo, fromIntegral hi)
 
--- | Choose a list of length `l` from a list allowing repetitions.
-chooseFromList :: Natural -> [a] -> Gen [a]
-chooseFromList l as = take l <$> shuffle (concatMap (replicate l) as)
+elementsRep :: Representable f => [a] -> Gen (f a)
+elementsRep xs = tabulate . const <$> elements xs
```
TurtlePU, line 107: This would generate a representable where all elements would be equal to each other, is this intended?

TurtlePU: (also -- should this belong to the Prelude?..)

vlasin: Thanks for noticing!

TurtlePU (approved):

PR #578, opened 2025-05-06
PlonkUp with updateable setup

vlasin:
Introducing the API to replace a Plonkup's public input with the corresponding constraint $x = c$. Updating the verifier's setup requires one scalar multiplication.


vlasin:
Adding everyone who is working on PlonkUp as reviewers because there are some minor changes to the protocol. Previously, public input constraints were the first `l` constraints. Now, the first `prvNum` constraints are "private inputs", then public inputs.

@vks4git @diS3e

symbolic-base/src/ZkFold/Protocol/Plonk/Prover.hs
```diff
@@ -197,7 +197,8 @@ plonkProve PlonkupProverSetup {..}
         h1_xi'  = h1X `evalPolyVec` (xi * omega)
         h2_xi   = h2X `evalPolyVec` xi
         lag1_xi = polyVecLagrange @_ @pv @(PlonkupPolyExtendedLength n) (value @n) 1 omega `evalPolyVec` xi
-        l_xi    = map (\i -> one // (scale n one * (xi - omega^i))) [1 :: Natural .. fromIntegral (length wPub)]
+        l1_xi   = one // (scale n one * (xi - omega))
```
TurtlePU, line 200: why not `finv (scale n (xi - omega))`, btw?

symbolic-base/src/ZkFold/Protocol/Plonkup/Input.hs
```diff
@@ -3,24 +3,21 @@
...
-newtype PlonkupInput l g = PlonkupInput { unPlonkupInput :: l (ScalarFieldOf g) }
+newtype PlonkupInput g = PlonkupInput { unPlonkupInput :: [ScalarFieldOf g] }
 
-instance (Show1 l, Show (ScalarFieldOf g)) => Show (PlonkupInput l g) where
+instance Show (ScalarFieldOf g) => Show (PlonkupInput g) where
     show (PlonkupInput v) = "Plonkup Input: " ++ show v
 
-instance (Arbitrary (l (ScalarFieldOf g))) => Arbitrary (PlonkupInput l g) where
+instance Arbitrary (ScalarFieldOf g) => Arbitrary (PlonkupInput g) where
     arbitrary = PlonkupInput <$> arbitrary
```
TurtlePU, line 20: This way, length of an input would be arbitrary, is this intended?

symbolic-base/src/ZkFold/Protocol/Plonkup/Update.hs
```diff
@@ -0,0 +1,90 @@
...
+    , KnownNat n
+    , AdditiveGroup g1
+    , Scale (ScalarFieldOf g1) g1
+    , UnivariateFieldPolyVec (ScalarFieldOf g1) pv
+    ) => PlonkupVerifierSetup i o n g1 g2 pv -> [ScalarFieldOf g1] -> [g1] -> PlonkupVerifierSetup i o n g1 g2 pv
+updateVerifierSetup setup@PlonkupVerifierSetup {..} inputs hs =
+    let
+        relation' = updateRelation relation inputs
+        PlonkupCircuitCommitments {..} = commitments
+        commitments' = commitments { cmQc = cmQc - sum (zipWith scale (toList inputs) (toList hs)) }
```
TurtlePU, line 88: `inputs` and `hs` are already lists, there's no need to call `toList`

symbolic-base/src/ZkFold/Protocol/Plonkup/Input.hs
vlasin: Actually, we don't need this instance since it is the output of the prover algorithm.

vlasin:
Ready for the second review. I also added `UtxoAccumulator` example, which we might use in some form for the bridge to the rollup.

TurtlePU (approved):

PR #580, opened 2025-05-14
Remove input functor from `Symbolic` circuit context

TurtlePU:
This PR splits `ArithmeticCircuit` into two datatypes:
* `CircuitContext a o`, which is commonly used as a Symbolic context;
* `ArithmeticCircuit a i o`, which is a wrapper around `CircuitContext` used for safe witness generation.

Thanks to the current interface of `SymbolicFold`, circuits produced using it are safe to evaluate even in presence in fold constraints; there would be a separate PR which would simplify the interface of `SymbolicFold` and change the `NewVar` accordingly to allow for implicit context switching.

Also the signatures are much nicer now, and the modules in `ArithmeticCircuit.*` are refactored a bit and cleaned up.


symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Context.hs
```diff
@@ -0,0 +1,358 @@
...
+            where
+                th = R.typeRep :: R.TypeRep h
+                tk = R.typeRep :: R.TypeRep k
+
+-- | Circuit context in the form of a system of polynomial constraints.
+data CircuitContext a o = CircuitContext
+    {
+        acSystem         :: Map ByteString (Constraint a),
+        -- ^ The system of polynomial constraints
+        acLookupFunction :: FunctionRegistry a,
```
vlasin, line 116: What is the purpose of this field?

symbolic-base/src/ZkFold/Symbolic/Compiler/ArithmeticCircuit/Context.hs
```diff
@@ -0,0 +1,358 @@
...
+-- It is a root hash (sha256) of a Merkle tree which is obtained from witness:
+--
+-- 1. Due to parametricity, the only operations inside witness are
+--    operations from 'WitnessField' interface;
+--
+-- 2. Thus witness can be viewed as an AST of a 'WitnessField' "language" where:
+--
+--     * leafs are 'fromConstant' calls and variables;
+--     * nodes are algebraic operations;
+--     * root is the witness value for new variable.
```
vlasin, line 344: Does it mean that we can serialize the witness computation algorithm?

vlasin:
@vks4git @diS3e Any comments?

diS3e (approved):

vks4git (approved):

vlasin (approved):

PR #584, opened 2025-05-16
Compilation to WASM

vks4git:
StdGen works differently in WASM for some reason. I had to replace random numbers with linear search in `roofOfUnity` and `getParams`. Feel free to suggest better approaches without using StdGen.


symbolic-base/symbolic-base.cabal
```diff
@@ -37,7 +37,9 @@ common options
       -Wredundant-constraints
       -Wunused-packages
       -O2
-      -threaded
+    if !arch(wasm32)
+      ghc-options:
+        -threaded
```
vks4git, line 42: wasm does not support threading

symbolic-examples/symbolic-examples.cabal
```diff
@@ -127,6 +134,10 @@ library
 
 benchmark bench-compiler-performance
     import:           options-exe
+    if arch(wasm32)
+      buildable: False
```
vks4git, line 138: executables mess with compilation to wasm

symbolic-examples/wasm/index.mjs
```diff
@@ -0,0 +1,78 @@
+import { WASI } from "https://cdn.jsdelivr.net/npm/@runno/wasi@0.7.0/dist/wasi.js";
```
vks4git, line 1: Just a demo

TurtlePU:
Maybe we can add building for different platforms (specifically WASM here) to CI checks to make sure that everything builds as intended?

symbolic-examples/symbolic-examples.cabal
TurtlePU: maybe this can be specified in `common options-exe` then?

symbolic-examples/symbolic-examples.cabal
```diff
@@ -35,7 +35,9 @@ common options
       -Wredundant-constraints
       -Wunused-packages
       -O2
-      -threaded
+    if !arch(wasm32)
+      ghc-options:
+        -O2
```
TurtlePU, line 40: should be `-threaded`, I guess

symbolic-examples/wasm/ghc_wasm_jsffi.js
```diff
@@ -0,0 +1,112 @@
+import { blake2b } from './blake2b.js'
```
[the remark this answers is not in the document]
TurtlePU: Is it possible to trigger the generation of this file during build?
In order to not store the generated output in the repo

symbolic-examples/wasm/proof.wasm
TurtlePU, line 1: Is it possible to trigger the generation of this file during build?
In order to not store generated output in the repo

symbolic-examples/compile_to_wasm.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+cabal clean
+
+# Compile to wasm
+cabal --with-compiler=wasm32-wasi-ghc-9.12 --with-hc-pkg=wasm32-wasi-ghc-pkg-9.12 --with-hsc2hs=wasm32-wasi-hsc2hs-9.12 build
```
TurtlePU, line 6: Instead of specifying `buildable: False` for executables in `.cabal`, I think you can build just a library here, like this:
```bash
cabal --with-compiler=... build symbolic-examples
```

symbolic-examples/compile_to_wasm.sh
TurtlePU, line 1: I hope in the future this could be cabal's own native functionality

symbolic-base/symbolic-base.cabal
```diff
@@ -230,11 +232,14 @@ library
...
+    if arch(javascript) || arch(wasm32)
       js-sources:
         js-src/blake2b.js
 
-    if arch(javascript)
+    if arch(wasm32)
+      build-depends:
+        ghc-experimental
+    if arch(javascript) || arch(wasm32)
       build-depends:
```
TurtlePU, line 243: Not a comment on the PR, but a comment on the future to deduplicate the declaration of dependencies in this file

symbolic-base/src/ZkFold/Protocol/Plonkup/Utils.hs
```diff
@@ -21,7 +19,7 @@ import           ZkFold.Prelude                     (iterateN', log2ceiling)
 import           ZkFold.Symbolic.Class              (Arithmetic)
 
 getParams :: forall a . (Ord a, FiniteField a) => Natural -> (a, a, a)
-getParams n = findK' $ mkStdGen 0
+getParams n = findK' 0
```
TurtlePU, line 22: You can use [`mkStdGen64`](https://hackage-content.haskell.org/package/random-1.3.1/docs/System-Random.html#v:mkStdGen64) here instead. My guess is that WASM is 32-bit architecture and, because of it, `mkStdGen` behaves differently (see docs for `mkStdGen64`)

symbolic-base/src/ZkFold/Protocol/Plonkup/Prover.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE CPP                 #-}
```
TurtlePU, line 2: Is this pragma needed here?

symbolic-examples/wasm/ghc_wasm_jsffi.js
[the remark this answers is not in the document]
TurtlePU: I consider the part with installing WASM target and running the build script to be a part of the example, so that we have a fully reproducible build. Storing the generated part makes it possible that some part of an example becomes outdated with respect to the GHC's WASM toolchain

symbolic-examples/wasm/blake2b.js
```diff
@@ -0,0 +1,443 @@
+
+
+// Blake2B in pure Javascript
+// Adapted from the reference implementation in RFC7693
+// Ported to Javascript by DC - https://github.com/dcposch
```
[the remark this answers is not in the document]
vks4git: Basically it's all the files from the original repo squashed into one for simpler JS FFI imports. We definitely don't want this in the original repo

symbolic-examples/wasm/ghc_wasm_jsffi.js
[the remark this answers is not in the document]
TurtlePU: @vks4git's wonderful `README` already documents all the steps required to reproduce the build!

symbolic-base/src/ZkFold/Protocol/NonInteractiveProof/Internal.hs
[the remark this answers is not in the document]
vks4git: Done

TurtlePU (approved):

Issue #296, opened 2024-10-13
Make `Conditional` part of `SymbolicData`


TurtlePU:
As defined currently, `Conditional` is a separate class with a single instance for types which are `SymbolicData` (under additional conditions). Expecting the addition of sum types in the future, it would be more convenient to consider `Conditional`-like functionality a part of the `SymbolicData` interface (this would also yield a more efficient version of conditional immediately):
```diff
-class Conditional b a where
- ...

-class SymbolicData d where
+class Symbolic (Context d) => SymbolicData d where
  ...
  pieces :: d -> Support d -> Context d (Layout d)
+ interpolate :: (Context d ~ c, Applicative f) => [(BaseField c, d)] -> c f -> f d
  restore :: (Support d -> Context d (Layout d)) -> d

+ -- in ZkFold/Symbolic/Data/Bool.hs
+ ifThenElse :: (SymbolicData d, Context d ~ c) => Bool c -> d -> d -> d
+ ifThenElse (Bool c) t e = unPar1 $ interpolate [(one, t), (zero, e)] c
```
I'd advise doing this after making `SymbolicData` derivable (#221) in order to avoid unnecessary diffs

TurtlePU:
> I'm not sure we should have Conditional be part of Symbolic because if we want to be able to generalize boolean logic across both Haskell and Symbolic as we are with the 3 other classes BoolType, Symbolic.Eq and Symbolic.Ord, then we don't want to require Symbolic. If on the other hand, we want to separate these interfaces then we should do so consistently for all those classes

@echatav, first of all, I want to integrate Conditional-like logic inside SymbolicData because:
* we can write it down for exactly the same types we can write down SymbolicData for, with exactly the same constraints;
* once we add sum types, we would need Conditional-like logic from the result type of a pattern match. And "result of a pattern match" as a separate type-level concept would complicate the interface IMO

That being said, completely agree that Symbolic.Eq and Symbolic.Ord have to be consistent! Something like
```haskell
class SymbolicData a => Eq a where
  (==) :: a -> a -> Bool (Context a)
class Eq a => Ord a where
  compare :: a -> a -> Tri (Context a)
```

With this in place, the "generalizing boolean logic" part can be completely offloaded to the Boolean library you were talking about the other day.

PR #600, opened 2025-05-27
Generalizing polynomials to require only `ZkFold.Data.Eq` instead of `GHC.Base.Eq`

TurtlePU:


symbolic-base/src/ZkFold/Protocol/Plonkup/Relation.hs
```diff
@@ -119,8 +122,17 @@ instance Semiring a => Semiring (Vector a) where
 
 instance Ring a => Ring (Vector a) where
 
+instance Conditional b a => Conditional b (Vector a) where
+    bool u v b = zipLongest (\x y -> bool x y b) u v
```
vlasin, line 126: Would it be better to put these instances inside `ZkFold.Data.Orphans`?

TurtlePU: Well, it's not an orphan...
But If you mean that `Conditional` instances shouldn't be written out by hand, I agree with you, we have a related issue #296. In my opinion, we should have only two instances:
```haskell
instance Conditional Bool a where bool = Haskell.bool -- in ZkFold.Data.Bool
instance (SymbolicData a, Context a ~ c) => Conditional (Bool c) a where bool = ... -- in ZkFold.Symbolic.Data.Bool
```
(And maybe one or two more in `ArithmeticCircuit.*`)

TurtlePU: I think it would actually be better to do one of the following
* introduce a newtype for "infinite sequences with finite nontrivial prefix" (as this is what these instances actually do)
* introduce an anonymous `KnownNat n` where `n` is a length of a lookup table and reuse instances from `ZkFold.Data.Vector n a` (but I'm not sure if GHC 9.6 has this)

vlasin (approved):

PR #607, opened 2025-06-06
Added `SimplePoly` newtype

TurtlePU:
Added a vector-backed `SimplePoly` newtype for use in `Symbolic` where previously `PolyVec` were used.
Also removed `isDiscrete` since it's not used anymore and breaks abstraction.

### DIFF
--- a/symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
+++ b/symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
@@ -2,10 +2,12 @@
 
 module ZkFold.Algebra.Polynomial.Univariate.Simple
     ( SimplePoly (coeffs)
+    , fromVector
     , toVector
     ) where
 
-import           Data.Function                        (($), (.))
+import qualified Data.Eq                              as Haskell
+import           Data.Function                        (on, ($), (.))
 import           Data.Functor                         (fmap)
 import           Data.List                            (zip)
 import           Data.Ord                             (min, (<))
@@ -16,6 +18,7 @@ import           Data.Vector                          (Vector)
 import qualified Data.Vector                          as V
 import qualified GHC.Num                              as Int
 import           Numeric.Natural                      (Natural)
+import           Text.Show                            (Show)
 
 import           ZkFold.Algebra.Class
 import           ZkFold.Algebra.Number                (KnownNat, integral)
@@ -27,7 +30,7 @@ import qualified ZkFold.Data.Vector                   as ZkFold
 --
 -- Not particularly efficient, but still usable in Symbolic
 -- since its operations do not require 'Prelude.Eq'.
-newtype SimplePoly a n = SimplePoly
+newtype SimplePoly a (n :: Natural) = SimplePoly
     { coeffs :: V.Vector a
       -- ^ Vector of coefficients in ascending-degree order.
       --
@@ -38,6 +41,15 @@ newtype SimplePoly a n = SimplePoly
       -- NOTE since 'SimplePoly' does not use 'Prelude.Eq', it is possible that
       -- there are some non-zeroed greater coefficients. Use with care.
     }
+    deriving (Show)
+
+instance
+    (AdditiveMonoid a, Haskell.Eq a, KnownNat n) =>
+    Haskell.Eq (SimplePoly a n) where
+    (==) = (Haskell.==) `on` toVector
+
+fromVector :: forall a n. ZkFold.Vector n a -> SimplePoly a n
+fromVector (ZkFold.Vector cs) = SimplePoly cs
 
 -- | Vector of exactly @n@ coefficients in ascending-degree order.
 toVector ::
--- a/symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
+++ b/symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
@@ -5,7 +5,7 @@ module ZkFold.Algebra.Polynomial.Univariate.Simple
     , toVector
     ) where
 
-import           Data.Function                        (flip, ($), (.))
+import           Data.Function                        (($), (.))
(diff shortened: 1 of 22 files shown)


symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+instance Ring a => UnivariateRingPolyVec a (SimplePoly a) where
+    SimplePoly p .*. SimplePoly q =
+        SimplePoly $ alignWith (uncurry (*) . T.fromThese zero zero) p q
+    (.*) = flip scale
+    (*.) = scale
+    (+.) c = SimplePoly . fmap (c +) . coeffs
+    (.+) = flip (+.)
+    toPolyVec :: forall n . KnownNat n => Vector a -> SimplePoly a n
+    toPolyVec = SimplePoly . V.take (integral @n)
+    fromPolyVec = coeffs
```
TurtlePU, line 101: @diS3e, is this implementation correct? Couldn't figure out since there's not enough documentation about these functions

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+    poly2vec = toPolyVec . fromPoly
+    vec2poly = toPoly . fromPolyVec
```
TurtlePU, line 103: I feel like for `poly2vec` and `vec2poly`, there's exactly one implementation thanks to their type: they have to work with arbitrary `Polynomial`s

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+    polyVecConstant = fromConstant
+    polyVecLinear k b = toPolyVec (V.fromList [b, k])
+    polyVecQuadratic a b c = toPolyVec (V.fromList [c, b, a])
```
TurtlePU, line 106: I don't know if `polyVecConstant`, `polyVecLinear` and `polyVecQuadratic` get any performance gain over `toPolyVec` in a generic implementation, maybe we can remove them from the class declaration as well

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+
+instance (Semiring a, KnownNat n) => Semiring (SimplePoly a n)
+
+instance (Ring a, KnownNat n) => Ring (SimplePoly a n)
+
+instance Ring a => UnivariateRingPolyVec a (SimplePoly a) where
+    SimplePoly p .*. SimplePoly q =
+        SimplePoly $ alignWith (uncurry (*) . T.fromThese zero zero) p q
+    (.*) = flip scale
+    (*.) = scale
```
TurtlePU, line 96: Since all our algebraic structures are commutative, I don't know if we need these two-sided aliases of `scale`. Also maybe these can be replaced with requirements of `Scale` instead.

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+    toPolyVec :: forall n . KnownNat n => Vector a -> SimplePoly a n
+    toPolyVec = SimplePoly . V.take (integral @n)
+    fromPolyVec = coeffs
+    poly2vec = toPolyVec . fromPoly
+    vec2poly = toPoly . fromPolyVec
+    polyVecConstant = fromConstant
+    polyVecLinear k b = toPolyVec (V.fromList [b, k])
+    polyVecQuadratic a b c = toPolyVec (V.fromList [c, b, a])
+    evalPolyVec (SimplePoly p) x =
+        sum [ c * x ^ i | (i, c) <- zip [(0 :: Natural)..] (V.toList p) ]
```
TurtlePU, line 108: I actually don't know if this or usual Horner scheme is going to be simpler in Symbolic, maybe they're equivalent...

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
```diff
@@ -0,0 +1,108 @@
...
+
+instance (Ring a, KnownNat n) => Ring (SimplePoly a n)
+
+instance Ring a => UnivariateRingPolyVec a (SimplePoly a) where
+    SimplePoly p .*. SimplePoly q =
+        SimplePoly $ alignWith (uncurry (*) . T.fromThese zero zero) p q
+    (.*) = flip scale
+    (*.) = scale
+    (+.) c = SimplePoly . fmap (c +) . coeffs
+    (.+) = flip (+.)
```
TurtlePU, line 98: Once again, these are equivalent since all our structures are commutative

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
vlasin: Yes, I think we can remove them from the class declaration. Those definitions are used in Plonkup, though. I propose to keep them outside the class.

symbolic-base/src/ZkFold/Algebra/Polynomial/Univariate/Simple.hs
diS3e: I think it is correct. But do you want to keep the parameter `n` and the vector length consistent? In that case, perhaps you can drop the `Eq` constraint for `UnivariateRingPolyVec`. Otherwise, you can additionally remove leading zeros

vks4git:
What do you think about adding some tests for SimplePoly?

vks4git (approved):

PR #612, opened 2025-06-10
`HashAlgorithm` refactor

TurtlePU:
Minor cleanup of signatures in IVC modules + a refactor of how random oracle extracts info. In particular:

* No overlapping instances;
* No "recursive hashing" problem, `oracle` is split into two phases &mdash; extraction of ring representation and hashing itself;
* No `AllowAmbiguousTypes` for `@algo` param, hashing function is passed explicitly;
* It is now possible to differentiate between hashing base field and base computation field if required.


symbolic-base/src/ZkFold/Protocol/IVC/Internal.hs
```diff
@@ -157,17 +156,19 @@ ivcProve f res witness =
...
 
-ivcVerify :: forall ctx0 ctx1 algo d k a i p c f . IVCAssumptions ctx0 ctx1 algo d k a i p c f
-    => StepFunction a i p
-    -> IVCResult k i c f
-    -> ((Vector k (c f), [f]), (Vector k (c f), c f))
-ivcVerify f res =
+ivcVerify ::
+    forall d k a i p c f . IVCAssumptions d k a i p c f =>
+    Hasher (FieldElement (CircuitContext a)) ->
+    Hasher f ->
```
TurtlePU, line 162: Unfortunately we have to pass two `Hasher`s here which work over two different base fields since there's no known bijection between them, and an invariant expected here is that they compute the same function.

Note that this problem was present before, but was obfuscated by type aliases for constraint sets!

However this can be solved by instead defining `type Hasher = forall f. C f => [f] -> f`, but I don't know which `C` to put here. `Ring` is enough for `mimcHash`, but I guess we would like to use other hashing algorithms as well.

symbolic-base/src/ZkFold/Protocol/IVC/Oracle.hs
```diff
@@ -1,75 +1,92 @@
...
+instance ( OracleSource a b, OracleSource a c, OracleSource a d
+         , OracleSource a e, OracleSource a f) => OracleSource a (b, c, d, e, f)
 
-class RandomOracle algo x a where
-    oracle :: x -> a
-    default oracle :: (Generic x, RandomOracle' algo (Rep x) a) => x -> a
-    oracle = oracle' @algo . from
+-- | A newtype to derive 'OracleSource' for any 'Foldable'
+-- as long as its element type is also 'OracleSource'.
+newtype FoldableSource f a = FoldableSource { foldableSource :: f a }
```
TurtlePU, line 47: I had to introduce this newtype to avoid overlapping (and **incoherent**, even) instances: `(a, b)` is `Foldable` in `b` as well, keep that in mind!

So actually this newtype is what we really need, because by using `FoldableSource f a` we signal that we drop any extra info in `f` which is not present via elements of `a`.

symbolic-base/src/ZkFold/Protocol/IVC/Predicate.hs
```diff
@@ -3,50 +3,31 @@
...
-    in Predicate {..}
+type StepFunction a i p =
+    CircuitContext a i -> CircuitContext a p -> CircuitContext a i
+
+predicate ::
+    forall a i p . (Arithmetic a, Binary a, LayoutFunctor i, LayoutFunctor p) =>
+    StepFunction a i p -> Predicate a i p
+predicate func = Predicate
+    { predicateEval = \x u ->
+        exec . compileWith solder (const (U1, U1)) $ func (embed x) (embed u)
```
TurtlePU, line 29: The "compiled" circuit here is constant, so the overhead in comparison with computation in `Interpreter` should be minor.

symbolic-examples/stats/MerkleTree.4
```diff
@@ -1,5 +1,5 @@
 MerkleTree.4
-Number of polynomial constraints: 4515
+Number of polynomial constraints: 5823
```
TurtlePU, line 2: Regression happened because previously constraints were not in PlonK form

symbolic-base/src/ZkFold/Protocol/IVC/Predicate.hs
vlasin: Never mind, it's not quite high-level as I see now.

symbolic-base/test/Tests/Protocol/IVC.hs
```diff
@@ -38,11 +44,16 @@ type D = 2
 type PARDEG = 5
 type PAR = Vector PARDEG F
 
+instance OracleSource F (C F) where
+    source _ = []
```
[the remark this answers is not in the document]
vlasin: Just in case, `Constant` here is just a mock. After we transition to `SymbolicData`, we can replace it with something more suitable.

vlasin (approved):

PR #619, opened 2025-06-17
Optimized `Conditional` instances

### LINKED ISSUE #296: Make `Conditional` part of `SymbolicData`

TurtlePU:
Now there's only two main `Conditional` instances:
```haskell
instance Conditional Haskell.Bool a
instance (SymbolicData a, Context a ~ c) => Conditional (Bool c) a
```
And the second one calls the optimized general `interpolate` function from `SymbolicData` class.

Can be viewed as both an improvement on its own (resolves #296) and as a part of a road towards sum types in Symbolic.


TurtlePU:
Division-by-zero exceptions in `divMod` for `UInt`, most likely has to do with laziness for `Conditional`, investigating

symbolic-base/src/ZkFold/Symbolic/Algorithm/Interpolation.hs
```diff
@@ -0,0 +1,62 @@
...
+  let ks =
+        [ product [pt - d | d <- ds] // product [c - d | d <- ds]
+        | (c, ds) <- cuts (fmap fst fs)
+        ]
+   in tabulate \r ->
+        -- \| branches to interpolate between
+        let is = (`index` r) . snd <$> toList fs
+         in sum [i * k | (i, k) <- P.zip is ks]
+
+interpolation
```
vlasin, line 41: Maybe call it `interpolate` in line with `interpolateW`?

symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
```diff
@@ -86,7 +88,18 @@ class
        , Payload x ~ GPayload (G.Rep x)
        )
     => x -> Support x -> Payload x (WitnessField (Context x))
-  payload x = gpayload (G.from x)
+  payload = gpayload . G.from
+
+  -- | Interpolates branch values between given points.
+  interpolate :: Context x ~ c => NonEmpty (BaseField c, x) -> c G.Par1 -> x
```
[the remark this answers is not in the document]
TurtlePU: We can indeed use interpolation for `c f` to implement this, but
* this approach is more efficient since it doesn't gobble circuits inside Symbolic datatype together
* this might turn out to not be redundant after future refactorings of `SymbolicData` class

symbolic-base/src/ZkFold/Symbolic/Algorithm/Interpolation.hs
TurtlePU: Done!

vlasin (approved):

PR #623, opened 2025-06-20
style(#611): use fourmolu

### LINKED ISSUE #611: Switch to fourmolu formatting in `symbolic`
- [ ] Add a config file
- [ ] Correct GitHub workflows
- [ ] Can we have fourmolu run on saving a file (like in Rust)? If not, let's add a formatting script.

sourabhxyz:
Closes #611.


symbolic-base/src/ZkFold/Algebra/EllipticCurve/BN254.hs
```diff
@@ -153,17 +161,78 @@ instance Finite BN254_GT where
...
-        , 1, 1, 0, 0, 0, 0,-1, 0, 1, 0, 0,-1, 0, 1, 1, 0
-        , 0, 1, 0, 0,-1, 1, 0, 0,-1, 0, 1, 0, 1, 0, 0, 0
-        ]
+  pairing p q =
+    BN254_GT $
+      finalExponentiation @Fr $
+        millerAlgorithmBN xi param p q
+   where
+    param =
+      [ 1
```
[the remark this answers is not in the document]
TurtlePU: Maybe there's a `FOURMOLU ignore` magic comment or something like that

fourmolu.yaml
```diff
@@ -0,0 +1,57 @@
...
+indent-wheres: false
+
+# Whether to leave a space before an opening record brace
+record-brace-space: true
+
+# Number of spaces between top-level declarations
+newlines-between-decls: 1
+
+# How to print Haddock comments (choices: single-line, multi-line, or multi-line-compact)
+haddock-style: multi-line
```
TurtlePU, line 30: If no one's against it, maybe we can have `single-line` here? 👉👈
It's closer to what we used before and plays better with dumb default neovim syntax highlighting

fourmolu.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+# How to print module docstring
+haddock-style-module: multi-line
```
TurtlePU, line 33: Same request here for `single-line`

symbolic-base/src/ZkFold/Algebra/EllipticCurve/BN254.hs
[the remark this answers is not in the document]
sourabhxyz: I'll wrap this block under `FOURMOLU_DISABLE`, see https://github.com/fourmolu/fourmolu?tab=readme-ov-file#magic-comments.

fourmolu.yaml
sourabhxyz: Have made a move to `single-line` 👍, though I personally prefer `multi-line` :)

symbolic-base/src/ZkFold/Symbolic/Algorithm/Hash/Blake2b/Constants.hs
```diff
@@ -32,23 +40,28 @@ blake2b_iv = fromList $ fromConstant @Natural <$>
...
+     , IsList (v (v Int))
+     , Item (v Int) ~ Int
+     , Item (v (v Int)) ~ v Int
+     )
+  => v (v Int)
+sigma =
+  fromList $
+    map
+      fromList
+      [ [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
```
sourabhxyz, line 55: I guess fourmolu does not format for nested lists, and thus it luckily didn't get affected.

.github/workflows/main-pull.yml
```diff
@@ -134,39 +139,28 @@ jobs:
...
-        cabal install stylish-haskell
-
-    - name: Lint Haskell
+    - name: Download and run fourmolu
       run: |
-        find ./ -name "*.hs" -exec sh -c 'for file do if ! grep -q -E -r "javascript_HOST_ARCH|LANGUAGE CPP|type \(\*\)" "$file"; then stylish-haskell --inplace "$file"; fi; done' sh {} \;
+        curl -L https://github.com/fourmolu/fourmolu/releases/download/v0.18.0.0/fourmolu-0.18.0.0-linux-x86_64 -o fourmolu
+        chmod +x fourmolu
+        git ls-files -z '*.hs' | xargs -P 12 -0 ./fourmolu --mode inplace
+        rm ./fourmolu
```
TurtlePU, line 158: Can we cache the binary maybe?

symbolic-base/src/ZkFold/Algebra/Class.hs
```diff
@@ -365,76 +363,75 @@ class (Ring a, SemiEuclidean a) => Euclidean a where
...
+  --   [Root property] If @rootOfUnity n == Just x@, @x ^ (2 ^ n) == one@
+  --   [Smallest root] If @rootOfUnity n == Just x@ and @m < n@, @x ^ (2 ^ m) /= one@
+  --   [All roots] If @rootOfUnity n == Just x@ and @m < n@, @rootOfUnity m /= Nothing@
+  rootOfUnity :: Natural -> Maybe a
+  rootOfUnity 0 = Just one
+  rootOfUnity _ = Nothing
 
 intPowF :: Field a => a -> Integer -> a
+
 -- | A default implementation for integer exponentiation. Uses only natural
```
TurtlePU, line 417: Wrong type of comment use here, either `-- ^` needed here or type declaration should go below the doc comment

fourmolu.yaml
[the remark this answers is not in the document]
sourabhxyz: Added!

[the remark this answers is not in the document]
sourabhxyz: Done! Using `by-scope`.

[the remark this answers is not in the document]
sourabhxyz: I'll disable `sort-constraints` as it's messing with out codebase since at places we have not given ordering of type parameters through `forall` and so it is implicitly defined by ordering of constraints.

fourmolu.yaml
```diff
@@ -0,0 +1,70 @@
...
+reexports: []
+
+# This option determines whether classes in deriving clauses should be sorted alphabetically or left alone.
+sort-derived-classes: true
+
+# This option determines whether deriving clauses should be sorted alphabetically or left alone.
+sort-derived-clauses: true
+
+# This option determines whether constraints should be sorted alphabetically or left alone.
+# We have set it to false for now since order of constraints lead to implicit ordering of type variables, changing it now would break existing code.
```
TurtlePU, line 69: A better practice would be to fix the order with `forall ... .` declaration, but I think it's better to be left to the separate PR since it changes semantics of the code

fourmolu.yaml
TurtlePU: Thank you so much c:

TurtlePU:
🔥 🔥 🔥

PR #624, opened 2025-06-23
draft: change RustData type to avoid copy

diS3e:


symbolic-base/rust-wrapper/src/utils.rs
```diff
@@ -40,54 +42,139 @@ cfg_if! {
...
+}
+
+pub unsafe fn drop<T>(b: Box<T>) {
+    let _ = Box::into_raw(b);
+}
+
+pub unsafe fn binary<T1, T2, R>(
+    a_ptr: *mut c_char,
+    b_ptr: *mut c_char,
+    f: fn(a: &T1, b: &T2) -> R,
```
TurtlePU, line 169: Why not just pass arguments here directly, as `a: T1` and `b: T2`?

diS3e: Because after calling function `f` i want to unwrap pointer from box. Otherwise Rust free this memory on exit function and on Haskell we will have dangling pointer on input params. For this unwrap function (`drop`) we need owning on input

symbolic-base/rust-wrapper/src/utils.rs
```diff
@@ -40,54 +42,139 @@ cfg_if! {
...
     let mut v = BigUint::from(scalar.into_bigint()).to_bytes_le();
     v.resize(std::mem::size_of::<ScalarField>(), 0);
     v
 }
+
+///
+/// # Safety
+/// The caller must ensure that valid pointers and sizes are passed.
+/// .
+unsafe fn h2r_ptr<T>(var: *const c_char, len: usize, pack: fn(&[u8]) -> T) -> *mut c_char {
```
TurtlePU, line 96: Why do these conversions still happen?

symbolic-base/src/ZkFold/FFI/Rust/Conversion.hs
```diff
@@ -49,6 +53,41 @@ instance {-# OVERLAPPABLE #-} (RustHaskell r h, Storable h) => Storable r where
...
+pointSize = sizeOf (undefined :: Rust_BLS12_381_G1_Point)
+
+scalarSize :: Int
+scalarSize = sizeOf (undefined :: Fr)
+
+o2nScalar :: Fr -> Fr
+o2nScalar old = unsafePerformIO $ do
+    withForeignPtr (rawData $ rawScalar old) $ \ptr -> do
+        ptrNew <- r_h2r_scalar ptr scalarSize
+        RScalar . RData <$> (newForeignPtr finalizerFree ptrNew )
```
TurtlePU, line 66: I think we need to run a finalizer which is different from standard `finalizerFree`, I don't know if this would work with Rust types

symbolic-base/rust-wrapper/src/poly.rs
```diff
@@ -1,188 +1,237 @@
...
-
-    let res = scalar_mul(l, r);
-
-    std::ptr::copy(res.as_ptr(), out as *mut u8, res.len());
+pub unsafe fn r_poly_mul(a_ptr: *mut c_char, b_ptr: *mut c_char) -> *mut c_char {
+    binary(
+        a_ptr,
+        b_ptr,
+        |a: &Vec<ScalarField>, b:   &Vec<ScalarField>| -> Vec<ScalarField> { 
+            let l = DensePolynomial::from_coefficients_vec(a.to_vec());
```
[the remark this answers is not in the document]
diS3e: No. I will need to separate a vector of scalars and a scalar polynomial in the `RustHaskell` typeclass. Already working on it.

symbolic-base/rust-wrapper/src/utils.rs
TurtlePU: Oh, I see, the `drop` below is not `std::mem::drop`, it's your own function. Maybe `Box::leak` is better to use instead of introducing the `drop<T>` function?

PR #627, opened 2025-06-24
Add `SymbolicFunction`

vlasin:


symbolic-base/src/ZkFold/Symbolic/Data/Class.hs
```diff
@@ -306,9 +262,26 @@ instance GSymbolicData f => GSymbolicData (G.M1 i c f) where
...
+
+type family Range a where
+  Range (x -> y) = Range y
+  Range y = y
+
+class (LayoutFunctor (Layout (Range f)), Domain (Range f) ~ Proxy (Context (Range f))) => SymbolicFunction f where
+  -- | Converts a function to a symbolic context.
+  apply :: f -> Domain f -> (Context (Range f)) (Layout (Range f))
+
+instance {-# INCOHERENT #-} SymbolicFunction y => SymbolicFunction (x -> y) where
```
TurtlePU, line 283: Maybe an `OVERLAPPING` would do instead of `INCOHERENT`?..

vlasin: Well, it requires changes to `Tests.Symbolic.Data.Common`.

TurtlePU: There are two possible approaches to try here:

1. Make `SymbolicFunction x` a superclass constraint for `SymbolicData x`. This would make GHC to defer the ambiguity check to the `instance SymbolicData` sites. Though I'm afraid this wouldn't work well with future refactoring of `SymbolicData` (maybe this would work with `QuantifiedConstraints`)

2. Add the `SymbolicFunction x` constraint to the functions. This would defer ambiguity check to function usage sites

TurtlePU: Though as I mentioned during the meeting, there's not going to be an overlap once we make `SymbolicData` into context functors :D. So I think the first approach should be the least intrusive

TurtlePU (approved):

PR #631, opened 2025-06-30
Fixed compiler performance benchmark

### LINKED ISSUE #552: Fix `bench-compilation-evaluation`
Currently, it takes hours to complete (I didn't wait for the `Rverse.32.3000` bench). Moreover, it displays numbers that are way too low (milliseconds). @TurtlePU suggested that it might reuse some computations between iterations.

TurtlePU:
Resolves #552.

For evaluation, my code was benchmarking the creation of the `NewVar -> a` function, not the actual evaluation. In the fix, a complete application of `eval :: AC a i o -> i a -> o a` is benchmarked instead.

For compilation, previous version of my code was creating a node in an evaluation graph per circuit which ended up being reused between benchmarks. In the fix, an `ExampleOutput` contains a function to compile, not the compiled circuit, so the compilation restarts correctly between runs.


TurtlePU:
`FAIL`ures because of division by zero in `divMod` implementation, investigating
Maybe worth it to split into separate task

PR #655, opened 2025-07-21
WIP Rust prover

vks4git:
First version of Rust prover.
Refactoring will be done after we make sure it works.


symbolic-base/rust-wrapper/src/prover.rs
```diff
@@ -0,0 +1,888 @@
...
+}
+
+fn zip_with<A, B, C, F>(a: &[A], b: &[B], mut f: F) -> Vec<C>
+where
+    F: FnMut(&A, &B) -> C,
+{
+    a.iter().zip(b.iter()).map(|(x, y)| f(x, y)).collect()
+}
+
+fn zip_with3<'a, A, B, C, D, F>(v1: &'a [A], v2: &'a [B], v3: &'a [C], f: F) -> Vec<D>
```
TurtlePU, line 314: `&`s have subtyping covariant in lifetimes, so you don't have to constrain arguments to have the same lifetime as far as I can see, so explicit lifetime annotation is not needed

symbolic-base/rust-wrapper/src/prover.rs
```diff
@@ -0,0 +1,888 @@
...
+    DensePolynomial::from_coefficients_vec(v)
+}
+
+fn rotL(pv: &DensePolynomial<ScalarField>) -> DensePolynomial<ScalarField> {
+    let mut v = pv.coeffs.clone();
+    v.rotate_left(1);
+    DensePolynomial::from_coefficients_vec(v)
+}
+
+trait Secret {
```
TurtlePU, line 391: Why do you need traits like this when you're going to have only single implementation? Why not define a function instead?

symbolic-base/rust-wrapper/src/prover.rs
```diff
@@ -0,0 +1,888 @@
...
+    fn get(&self, ix: usize) -> ScalarField;
+}
+
+impl Secret for PlonkupProverSecret {
+    fn get(&self, ix: usize) -> ScalarField {
+        self.secret[ix - 1]
+    }
+}
+
+trait Elementwise<T> {
```
TurtlePU, line 401: Looks like `T` has to be an associated type rather than a parameter here

symbolic-base/rust-wrapper/src/prover.rs
```diff
@@ -0,0 +1,888 @@
...
+    fn elementwise<F>(&self, mut f: F) -> Self
+    where
+        F: FnMut(&T) -> T,
+    {
+        let new_coeffs = self.coeffs.iter().map(|c| f(c)).collect();
+        DensePolynomial::from_coefficients_vec(new_coeffs)
+    }
+}
+
+pub trait ZipWith<T> {
```
TurtlePU, line 417: Same here -- looks like `T` is actually an associated type

symbolic-base/rust-wrapper/src/scale.rs
```diff
@@ -14,7 +13,7 @@ use crate::utils::{
 
 // This is inner function from arkmsm crate, derived from benchmark results. May not be optimal for all configurations
 // https://github.com/snarkify/arkmsm/blob/main/src/msm.rs
-const fn get_opt_window_size(k: u32) -> u32 {
+pub const fn get_opt_window_size(k: u32) -> u32 {
```
TurtlePU, line 16: Is this function used anymore?

TurtlePU (dismissed):

vlasin (dismissed):

PR #671, opened 2025-07-29
Vks4git/server side proving

vks4git:
Bugfixes for the smart wallet backend and server-side proving


vlasin:
Need to override the bench, I guess.

PR #682, opened 2025-08-11
Integrating `OperationRecord` into IVC

TurtlePU:
* strengthened `HomomorphicCommit` to be an additive action rather than just a function on curve points;
* split commitment type into `pt` type which is an additive group and a `c` type which can be additively acted upon by `pt`.
* made `OperationRecord` to store a proper list of operations to check

### DIFF
--- a/symbolic-base/src/ZkFold/Algebra/Class.hs
+++ b/symbolic-base/src/ZkFold/Algebra/Class.hs
@@ -248,16 +248,21 @@ class FromConstant a a => AdditiveSemigroup a where
   double :: a -> a
   double a = a + a
 
+-- | A class of types with a neutral element with respect to addition.
+-- Split into a separate class since defining addition might need additional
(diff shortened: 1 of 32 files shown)


symbolic-base/src/ZkFold/Algebra/Class.hs
```diff
@@ -248,16 +248,21 @@ class FromConstant a a => AdditiveSemigroup a where
   double :: a -> a
   double a = a + a
 
+-- | A class of types with a neutral element with respect to addition.
+-- Split into a separate class since defining addition might need additional
+-- (undesirable) constraints.
+class Zero a where
```
vks4git, line 254: Why was it necessary to separate this class from `AdditiveMonoid`?
And perhaps we need to separate `One` from `MultiplicativeMonoid` as well for consistency?

PR #690, opened 2025-09-03
feat(#689, symbolic-ledger): pivot to simplified account model

sourabhxyz:


symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/Transaction/Batch.hs
```diff
@@ -8,45 +8,33 @@ import GHC.Generics (Generic)
...
-  -- ^ Hash of 'TransactionBatchData' indexed by the corresponding data availability source.
-  , tbBridgeIn :: HashSimple context
-  -- ^ Hash of the 'AssetValues' that are bridged into the ledger.
-  , tbBridgeOut :: HashSimple context
-  -- ^ Hash of the 'AssetValues' that are bridged out of the ledger.
-  , tbValidityInterval :: Interval context
-  -- ^ The validity interval of the transaction batch. The bounds are inclusive.
-  , tbPreviousBatch :: HashSimple context
-  -- ^ Hash of the previous transaction batch.
+data TransactionBatch c = TransactionBatch
```
vlasin, line 20: This could be simply a vector of transactions.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/State.hs
```diff
@@ -0,0 +1,46 @@
...
+import ZkFold.Symbolic.Ledger.Types.Nonce (KnownRegistersNonce, Nonce)
+import ZkFold.Symbolic.Ledger.Types.Root (Root)
+import ZkFold.Symbolic.Ledger.Types.Value (AssetValues, KnownRegistersAssetQuantity)
+
+-- | Defines the on-chain representation of the Symbolic Ledger state transition.
+data State context = State
+  { sPreviousStateHash :: HashSimple context
+  -- ^ Hash of the previous state.
+  , sBridgeIn :: HashSimple context
+  -- ^ Hash of the 'AssetValues' that are bridged into the ledger.
```
vlasin, line 24: More precisely, it should be a hash of a fixed-size vector of pairs `(Address, AssetValue)`.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Types/State.hs
```diff
@@ -0,0 +1,46 @@
...
+  , sBridgeOut :: HashSimple context
+  -- ^ Hash of the 'AssetValues' that are bridged out of the ledger.
+  , sDataHash :: HashSimple context
```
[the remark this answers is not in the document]
sourabhxyz: Could it be that more than one possible application lead to same state update? In that case, it might be worth committing to the action which led to this update.

PR #691, opened 2025-09-04
Symbolic Merkle tree

### LINKED ISSUE #687: Add Symbolic Merkle tree

TurtlePU:
Resolves #687.

### DIFF
--- a/symbolic-base/src/ZkFold/Data/Orphans.hs
+++ b/symbolic-base/src/ZkFold/Data/Orphans.hs
@@ -70,6 +70,9 @@ instance Arbitrary a => Arbitrary (Par1 a) where
 instance (Arbitrary (f a), Arbitrary (g a)) => Arbitrary ((f :*: g) a) where
   arbitrary = (:*:) <$> arbitrary <*> arbitrary
 
+instance Arbitrary (f (g a)) => Arbitrary ((f :.: g) a) where
+  arbitrary = Comp1 <$> arbitrary
+
 instance ToJSON a => ToJSON (U1 a)
 
 instance FromJSON a => FromJSON (U1 a)


symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -0,0 +1,186 @@
...
+---------------------------- conversion functions ------------------------------
+
+toBaseHash :: Symbolic c => FieldElement c -> WitnessField c
+toBaseHash = unPar1 . witnessF . fromFieldElement
+
+toBaseLeaves :: Payloaded (Leaves d) FieldElement c -> Leaves d (WitnessField c)
+toBaseLeaves = fmap (unPar1 . fst) . runPayloaded
+
+fieldElemW :: Symbolic c => WitnessField c -> FieldElement c
+fieldElemW = FieldElement . embedW . Par1
```
vlasin, line 179: `fromBaseHash`?

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -0,0 +1,228 @@
...
+import ZkFold.Symbolic.Data.Maybe (Maybe, fromJust, guard)
+import ZkFold.Symbolic.Data.Payloaded (Payloaded (..), payloaded, restored)
+import ZkFold.Symbolic.MonadCircuit (IntegralOf, toIntegral)
+import ZkFold.Symbolic.WitnessContext (WitnessContext (..))
+
+type Leaves d = Vector (Base.MerkleTreeSize d)
+
+data MerkleTree d a c = MerkleTree
+  { rootHash :: FieldElement c
+  , leafData :: Payloaded (Leaves d) a c
```
sourabhxyz, line 53: I had a glance at this PR last week, and just wanted to mention that for usage in rollup, we won't have access to data for all leaves. I hope the code allows for it.

cc @vlasin

vlasin: > I had a glance at this PR last week, and just wanted to mention that for usage in rollup, we won't have access to data for all leaves. I hope the code allows for it.

Yeah, good point.

vlasin:
Comments:
1. I suggest using the same naming conventions in the base module and moving the definition of `Leaves d` there.
2. It would be great if we had a clear indication of which functions do not produce constraints (`fromLeavesW`).

vlasin (approved):

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
sourabhxyz, line 1: It seems that we don't support use case where we don't know all leaves of merkle tree. I was looking at `replace` function, and usually replacement should be taking merkle proof of previous element and likewise to insert new element, but apparently, we are requiring all elements (leaves) of tree to do replacement 🤔

sourabhxyz: > How is one going to construct the original merkle proof in the first place if all of the leaves of merkle tree are not known?

In our use case, we have a merkle tree which is private to the user, so they know all the leaves (and can thus provide proofs) but not necessarily others.

TurtlePU: @sourabhxyz, is it ok with you if I add the desired functionality in the next PR? I have another PR which depends on this one and is ready for review so I don't want to stall that one for too long

sourabhxyz: Yes, let's please proceed with merging this PR, thanks!

Issue #659, opened 2025-07-21
Unify `Payload` and `Layout` in a single concept


TurtlePU:
After #635 is done, the only incongruency left in `SymbolicData` is a `Layout`/`Payload` split. Arguably, a `SymbolicData` should decide itself what is constrained and what is not; the only problem are some of the algorithms we have which rely on `Layout`'s finiteness which would not be guaranteed when it will be merged with `Payload`. So the task is to somehow change these algorithms to not rely on `Layout` explicitly; either `interpolate` or some kind of `footprint :: (Symbolic c, SymbolicData d) => d c -> [FieldElement c]` should do.

Issue #658, opened 2025-07-21
Create a lower-kinded `SymbolicData` counterpart


TurtlePU:
In order to migrate to lower-kinded Symbolic DSL, we have to introduce a modified version of `SymbolicData` (let's call it `SD'` here for brevity) along with an adapter which turns `SymbolicData` in an adapter context (see #657) into an `SD'`. Might be worth resolving #635 first for easier definition and later migration.

PR #709, opened 2025-09-24
New Symbolic(Data) classes

### LINKED ISSUE #657: Add lower-kinded counterpart to `Symbolic` class
With an introduction of new backend in #654, contexts can be thought of as special `Field`-like `Type`s instead of `(Type -> Type) -> Type` functors. The task here is to define a typeclass (let's call it `Symbolic'` for now) which describes this functionality along with an adapter lifting instances of `Symbolic'` to `Symbolic` we have now for easier migration.

TurtlePU:
Resolves #657, #658 & #659 by introducing new Symbolic & SymbolicData classes along with adapters between old and new interface defined in `ZkFold.Symbolic.Compat`.
Experimental backend doesn't _yet_ implement new interface because of a discovered possibility to lose constraints in its current implementation, will open a new issue to resolve this.

### DIFF
--- a/symbolic-base/src/ZkFold/Symbolic/Data/V2.hs
+++ b/symbolic-base/src/ZkFold/Symbolic/Data/V2.hs
@@ -34,12 +34,12 @@ class (forall c. HasRep f c => RepData f c) => SymbolicData (f :: Type -> Type) 
   type HasRep f (c :: Type) :: Constraint
   type HasRep f c = HasRep (G.Rep1 f) c
 
-  arithmetize :: Symbolic c => f c -> Layout f c c
-  default arithmetize
+  toLayout :: Symbolic c => f c -> Layout f c c
+  default toLayout
     :: (Symbolic c, G.Generic1 f, SymbolicData (G.Rep1 f))
     => Layout f c ~ Layout (G.Rep1 f) c
     => f c -> Layout f c c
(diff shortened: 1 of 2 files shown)


symbolic-base/src/ZkFold/Symbolic/Data/V2.hs
```diff
@@ -0,0 +1,95 @@
...
+    :: (Symbolic c, G.Generic1 f, SymbolicData (G.Rep1 f))
+    => c -> NonEmpty (Natural, f c) -> f c
+  interpolate v = G.to1 . interpolate v . fmap (fmap G.from1)
+
+  restore :: Symbolic c => Layout f c c -> f c
+  default restore
+    :: (Symbolic c, G.Generic1 f, SymbolicData (G.Rep1 f))
+    => Layout f c ~ Layout (G.Rep1 f) c
+    => Layout f c c -> f c
+  restore = G.to1 . restore
```
vlasin, line 55: I am thinking if there are better names for `arithmetize` and `restore`?

TurtlePU: I like the original names so I used them here, but sure, why not, we can rename it to something else.
The meaning of these operations is repacking the data from the layout, so maybe `toLayout`/`fromLayout`?

vlasin (approved):

PR #712, opened 2025-09-26
Signature check for ledger

sourabhxyz:


symbolic-base/src/ZkFold/Symbolic/Data/FFA.hs
```diff
@@ -5,7 +5,7 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE NoStarIsType #-}
 
-module ZkFold.Symbolic.Data.FFA (UIntFFA (..), FFA (..), KnownFFA, FFAMaxBits, toUInt, fromInt) where
+module ZkFold.Symbolic.Data.FFA (UIntFFA (..), FFA (..), KnownFFA, FFAMaxBits, toUInt, fromInt, fromUInt) where
```
sourabhxyz, line 8: I am using `fromUInt` in my code, I hope that's fine?

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Validation/Transaction.hs
```diff
@@ -162,14 +170,29 @@ validateTransaction utxoTree bridgedOutOutputs tx txw =
...
                   && (acc `MerkleTree.contains` merkleEntry)
+                  && Poseidon.hash publicKey
+                  == utxo.uOutput.oAddress
+                  && eddsaVerify
+                    ( \rPoint' publicKey' m ->
+                        let Elliptic.AffinePoint rPointx rPointy = affinePoint rPoint'
+                            Elliptic.AffinePoint publicKeyx publicKeyy = affinePoint publicKey'
+                         in fromUInt
+                              ( toUInt
+                                  (Poseidon.poseidonHashDefault [rPointx, rPointy, publicKeyx, publicKeyy, fromUInt . from $ m])
```
vlasin, line 189: I think it would be nice if we could do `poseidonHash (rPoint, publicKey, m)` instead of this expression (we can define hashing for an EC point as hashing of its coordinates). Can we do that?

vlasin: We could just use tuples in a situation like this, no?

sourabhxyz: @vlasin Yes, that's a good suggestion! Shall I have this change in my offchain PR since it is based upon this branch?

PR #725, opened 2025-10-16
feat(#724): hand-crafted test for updateLedgerState

sourabhxyz:


symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -130,6 +131,8 @@ data MerkleEntry d c = MerkleEntry
   }
   deriving (Generic1, SymbolicData, SymbolicInput)
 
+deriving stock instance HShow c => P.Show (MerkleEntry d c)
```
sourabhxyz: https://github.com/zkFold/symbolic/pull/725/files#r2460571015

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Validation/Transaction.hs
```diff
@@ -206,6 +208,7 @@ validateTransaction utxoTree bridgedOutOutputs tx txw =
                   :*: (outputIx + one)
                   :*: ( outsValidAcc
                           && foldl' (\found boutput -> found || output == boutput) false (unComp1 bridgedOutOutputs)
+                          && (output /= nullOutput)
```
sourabhxyz: Likely this is not needed, but it's not useful to have a "null" bridged-out output.

PR #741, opened 2025-11-05
feat(#733): compilation of rollup circuit

### LINKED ISSUE #733: Method to generate setup bytes and proof bytes for rollup circuit

sourabhxyz:
Closes #733.

### DIFF
--- a/symbolic-ledger/src/ZkFold/Symbolic/Ledger/Circuit/Compile.hs
+++ b/symbolic-ledger/src/ZkFold/Symbolic/Ledger/Circuit/Compile.hs
@@ -1,6 +1,13 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 
 module ZkFold.Symbolic.Ledger.Circuit.Compile (
+  LedgerContractInput (..),
+  LedgerContractOutput,
+  LedgerCircuitGates,
+  LedgerCircuit,
+  PlonkupTs,
+  TranscriptConstraints,
+  ledgerContract,
   ledgerCircuit,
   ledgerSetup,
   ledgerProof,


symbolic-base/src/ZkFold/Algebra/EllipticCurve/Class.hs
sourabhxyz: @TurtlePU Could you take a look if these instances are fine?

symbolic-base/src/ZkFold/Symbolic/Data/MerkleTree.hs
```diff
@@ -4,7 +4,7 @@
 {-# LANGUAGE UndecidableInstances #-}
 
 module ZkFold.Symbolic.Data.MerkleTree (
-  MerkleTree,
+  MerkleTree (mHash),
```
sourabhxyz: @TurtlePU I exported `mHash` to get root hash of the Merkle tree, I hope it's fine.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Circuit/Compile.hs
```diff
@@ -0,0 +1,351 @@
...
+  , z2_xi'_int :: !Integer
+  , h1_xi'_int :: !Integer
+  , h2_xi_int :: !Integer
+  , l_xi :: ![ZKF]
+  , l1_xi :: !ZKF
+  }
+  deriving stock (Generic, Show)
+  deriving anyclass (FromJSON, ToJSON)
+
+mkSetup :: forall i n. KnownNat n => SetupVerify (PlonkupTs i n ByteString) -> ZKSetupBytes
```
sourabhxyz: Though `mkSetup` and `mkProof` are copy-pasted from `ZkFold.Symbolic.Examples.SmartWallet`, they cannot however be imported from that module since we have defined `PlonkupTs` _differently_. Essentially, we could have only borrowed types like `ZKProofBytes` from it which for now I thought not to do so as to avoid dependency upon `symbolic-examples`. We could however move these types to `symbolic-base` if we deem it generally useful.

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Circuit/Compile.hs
```diff
@@ -0,0 +1,351 @@
...
+  )
+    :*: ( sPreviousStateHash lciNewState
+            :*: (mHash . sUTxO $ lciNewState)
+            :*: sLength lciNewState
+            :*: (hHash . sBridgeIn $ lciNewState)
+            :*: (hHash . sBridgeOut $ lciNewState)
+        )
+    :*: validateStateUpdate lciPreviousState lciTransactionBatch lciNewState lciStateWitness
+
+-- TODO: Is this circuit gate count enough?
```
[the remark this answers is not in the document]
vlasin: Try increasing it. If the result is the same, then it's enough :)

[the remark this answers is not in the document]
vlasin: You can compile and output the constraint number.

vlasin (approved):

symbolic-ledger/src/ZkFold/Symbolic/Ledger/Circuit/Compile.hs
sourabhxyz: It seems that we are fine with moving it to `symbolic-base`. I'll do it in a follow up PR 👍

PR #803, opened 2026-02-18
Plonk fix

vks4git:
Fixes and speeds up Plonk proofs.

The function `ZkFold.Protocol.Plonkup.PlonkConstraint.toPlonkConstraint` replaced polynomials which it could not parse with just zero constraints. In some cases, up to 70% of the circuit ended up being dummy constraints.

That functions was also slow. The function was generating redundant permutations while trying to find coefficients. I've replaced this function with another one that retrieves the coefficients diretly without any permutations. Here's the new flamegraph. Now the bottleneck is Rust prover (specifically, serialisation).

I've also introduced checks into the `MonadCircuit` instance. Now it should be impossible to create polynomial constraints that can not be represented as Plonk constraints.

Current flamegraph:

image


sourabhxyz (dismissed):
Good!

sourabhxyz (approved):

